Issue #47, opened 2021-02-22
Think about changing name of package to match name of dir


imbstack:
https://github.com/mozilla-it/ctms-api/pull/45#discussion_r579536407 is the context here

bsieber-mozilla:
Closing issue

PR #81, opened 2021-03-02
Issue #47, if desired this PR addresses this issue

bsieber-mozilla:
## Proposed changes

Update TOML to align package to be named CTMS instead of ctms-api, which also has poetry to look at CTMS directory instead of using the packages section of the toml to determine where to look.

Not a necessary change, but just came up per the issue.

## Types of changes

- [x] Documentation Update (if none of the other choices apply)

## Checklist

- [x] I have read the [guides](https://github.com/mozilla-it/ctms-api/tree/main/guides)
- [x] I have followed the [Mozilla Lean Data Policies](https://www.mozilla.org/en-US/about/policy/lean-data/)

## Further comments

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,7 @@ RUN curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poet
 # and install only runtime deps using poetry
 WORKDIR $PYSETUP_PATH
 COPY ./poetry.lock ./pyproject.toml ./
-RUN poetry install --no-dev --no-root
+RUN poetry install --no-dev
 
 
 # 'development' stage installs all dev deps and can be used to develop code.
@@ -51,7 +51,7 @@ RUN chmod +x /docker-entrypoint.sh
 
 # venv already has runtime deps installed we get a quicker install
 WORKDIR $PYSETUP_PATH
-RUN poetry install --no-root
+RUN poetry install
 
 WORKDIR /app
 COPY . .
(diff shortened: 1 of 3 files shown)


jwhitlock (approved):
I've gotten a little more comfortable with this change as I work with installing new dependencies, approving.

I also think the "right" think to do is to install the project in the ``development`` build of the ``Dockerfile``, but that should be a different PR.

bsieber-mozilla:
Yeah, I think @imbstack had run into an issue with the use of the `packages` section in the pyproj so it was removed [here](https://github.com/mozilla-it/ctms-api/commit/6c15930a10fbc6f118247ba83cae2e995d4a7fe1) and links to a poetry issue addressing the use of --no-root, when a `packages` section exists (it's a indexing issue it appears).

However, we can likely roll those, previously necessary changes of --no-root, back when we merge this PR in (or I can add it in here).

What are your thoughts @imbstack? Maybe there was another issue that popped up that I am unaware of that required the change, but I wouldn't want to slyly re-add something you were fighting. 

We can:

- [ ] drop this PR (it's not adding functionality, it's adhering to the poetry naming conventions of the package, w/o needing the packages section)
- [ ] add a commit to this PR addressing the --no-root
- [ ] add a followup PR with the --no-root

imbstack (approved):
awesome! feel free to address the `--no-root` thing in this PR or a followup. I agree this makes it unnecessary to have

bsieber-mozilla:
I think this was the previous method that was called out, with the `--no-root` removal we have the poetry install and shell, able to use the internal packages, without `python` prefixed

Screen Shot 2021-03-08 at 2 41 07 PM

bsieber-mozilla:
Merging this in; let me know of any concerns and we can rb

PR #83, opened 2021-03-04
Convert docker-compose to use environment files

jwhitlock:
Environment files like [.env](https://docs.docker.com/compose/environment-variables/#the-env-file) are considered after the environment settings in``docker-compose.yml`` and the shell. The ``docker/config/local_dev.env`` file has settings for the local development environment, such as the database URL. The ``.env`` file has developer-specific settings, is created when a ``make`` command is used, and is ignored by ``git``.

To demonstrate  possibilities, I've added a few override for ``Makefile`` behavior:

* ``MK_WITH_SERVICE_PORTS=--service-ports`` - Add ``--service-ports`` to calls to ``docker-compose run``, to make it more likely that postgres ports will stay open
* ``MK_KEEP_DOCKER_UP=1`` - Avoid calling ``docker-compose down`` after ``make test``

The real use cases are in the future - adding a dev value for the per-deployment ``CTMS_SECRET_KEY`` for crypto operations, adding a switch for development vs live servers, etc. If it doesn't make sense to do it now, I can make the change as part of the OAuth2 stuff, when it will make more sense.

This is based on [similar functionality in MLS](https://github.com/mozilla/ichnaea/tree/main/docker/config), which doesn't use ``docker-compose`` for deployment. I have no experience with that use case, so I'm not sure if this helps or hurts.

### DIFF
--- a/docker/config/env.dist
+++ b/docker/config/env.dist
@@ -1,0 +1,9 @@
+# This file is for settings that are specific to your docker local development
+# environment. Since these are specific to your local development environment,
+# don't check them in.
+
+# Set this to always call docker-compose run with --service-ports in the Makefile
+# MK_WITH_SERVICE_PORTS=--service-ports
+
+# Set this to skip calling "docker-compose down" after "make tests"
+# MK_KEEP_DOCKER_UP=1
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ reports/
 htmlcov/
 .isort.cfg
 docs/_build
+.env
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+# Include .env and export it so variables are available in the Makefile.
+include .env
+export
+
+
 .PHONY: help
 help:
 	@echo "Usage: make RULE"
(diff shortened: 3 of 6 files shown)


cvalaas (approved):

docker/config/env.dist
```diff
@@ -0,0 +1,9 @@
+# This file is for settings that are specific to your docker local development
```
imbstack, line 1: what if we just put all `.env` in this dir into gitignore and wrote these docs in a README in there to help avoid accidental commitment?

imbstack (approved):
makes sense to me!

imbstack:
I think there's a couple additional steps to do to get fastapi to read .env files iirc?

docker/config/env.dist
jwhitlock: To be clear, this message is meant for the ``.env`` file that gets created from ``env.dist``. The idea is that everything in this file is disabled by default, and someday, if and when a dev needs it, they can go into their ``.env`` and selectively change things. However, as we add things, they go into ``env.dist``, so future devs get those new potential settings. It's not ideal, but simple.

This approach isn't risk free. On one project I've been on, the SRE used these files to populate the production environment, rather than as a template for the prod environment, and copied over some stuff that was meant for development only. I'm not sure how to avoid that entirely, but not checking in the file would help. Comments come in as a distant second as a mitigation strategy.

When an environment file is mentioned in ``docker-compose``, it must be present or ``docker-compose`` fails. If these weren't present or auto-created, then step 1 for the project would be to read some docs about creating a ``.env`` file, with all the options you could put in there.  I think that puts a lot of complexity on a developer before they have any experience with a project. We *might* have to create the ``.env`` file anyway, for CI to work, so there would be an auto-created ``.env`` hidden in the code somewhere.

If I misunderstood your question, maybe a PR would help?

jwhitlock:
> I think there's a couple additional steps to do to get fastapi to read .env files iirc?

I mostly using ``make test``, ``make start``,  and ``make shell``, to run things in ``docker`` through ``docker-compose``. These are added as environment variables, and FastAPI is picking up those. I suppose you'll need something like ``source .env docker/config/local_dev.env`` if you are trying to run things on your host machine.

PR #86, opened 2021-03-10
Adding a staging tag to the commits on main when docker builds them

duallain:
main will add stg- to tags
prod will use releases, but nothing in this pr helps make that real, just documented for the future
both of these can then be matched with regex in the helm release

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,22 +77,6 @@ The project is structured with the following in mind:
     - pytest tests
 
 ---
-## Deployment
-
-We use a variety of technologies to get this code into production.  Starting from this repo and going outwards:
-
-1. github actions builds and deploys a docker container to ecr
-    1. prs and pushes to this repo will build and push a 'short-sha' container to AWS' ECR
-    1. Code merged to main will trigger a build that prefixes the short sha with literal 'stg-'
-    1. Code merged to prod will trigger a build that prefixes the short sha with literal 'prod-'
-1. A helm release is configured in ctms-infra 
-    1. https://github.com/mozilla-it/ctms-infra/tree/main/k8s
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -76,6 +76,22 @@ The project is structured with the following in mind:
 - tests/unit/
     - pytest tests
 
+---
+## Deployment
```
jwhitlock, line 80: I think this belongs in ``guides/deployment_guide.md``, which is outdated.

README.md
```diff
@@ -76,6 +76,22 @@ The project is structured with the following in mind:
...
+
+We use a variety of technologies to get this code into production.  Starting from this repo and going outwards:
+
+1. github actions builds and deploys a docker container to ecr
+    1. prs and pushes to this repo will build and push a 'short-sha' container to AWS' ECR
+    1. Code merged to main will trigger a build that prefixes the short sha with literal 'stg-'
+    1. Code merged to prod will trigger a build that prefixes the short sha with literal 'prod-'
```
jwhitlock, line 87: Can we use git tags to identify production code instead of a branch ``prod``?

If we need to use a branch, then we may need some GitHub settings to protect that branch.

jwhitlock (commented):
Thanks @duallain! This looks like a decent change to only push merges to ``main``.

It looks like the build is broken, but I can't determine from these changes why. Any ideas?

jwhitlock (approved):
Thanks for the changes @duallain!

It looks like that build step may not work if the PR is from a fork rather than a branch on the mozilla-it/ctms-api repo. That seems like the right thing - PRs from forks of a public repo should be trusted less. I'd prefer something other than a build step failure - exit early, or load safe alternate credentials - but that is a different issue for a new PR.

Issue #89, opened 2021-03-10
Implement mozilla-services/dockerflow


jwhitlock:
https://github.com/mozilla-services/dockerflow is a specification for automated building, testing and publishing of docker web application images that comply to a common set of behaviors. 

Tracked in Jira:OSS-278

Things we do already (fix if I'm wrong):

* Trigger builds through pull requests and merge to the ``main`` branch.
* Accept configuration through environment variables.
* Send text logs to `stdout` or `stderr`.
* Serve its own [static content](https://github.com/mozilla-services/Dockerflow/blob/main/docs/serving-static-content.md).
* Dockerfile has both ENTRYPOINT and CMD instructions to start the service.
* Following most of the [Guidelines for Building a Container](https://github.com/mozilla-services/Dockerflow/blob/main/docs/building-container.md), with maybe bonus points for multi-stage builds.

Changes to make:

* [x] - Listen at environment variable ``$PORT`` (8000 recommended) rather than hard-coded port 80
* [x] - Create a [JSON version object](https://github.com/mozilla-services/Dockerflow/blob/main/docs/version_object.md) at `/app/version.json`
* [x] - Respond to `/__version__` with the contents of `/app/version.json`.
* [x] - Respond to `/__heartbeat__` with a HTTP 200 or 5xx on error. This should check backing services like a database for connectivity and may respond with the status of backing services and application components as a JSON payload.
* [x] - Respond to `/__lbheartbeat__` with an HTTP 200. This is for load balancer checks and should **not** check backing services (close to existing ``/health`` endpoint)
* [x] - In Dockerfile, create ``app`` group and user, GID 10001 and UID 10001, set as user (currently ``root``)

Optional changes:
* [x] Output to stdout in [mozlog](https://wiki.mozilla.org/Firefox/Services/Logging) JSON schema. There are ways to use this in production but a friendlier human format in development.

Maybe changes to make, need to ask @duallain or @sciurus :

* [ ] Dockerflow says containers should be published to Docker Hub. We publish to Amazon ECR. Do we need to change docker repositories?
* [ ] Dockerflow says CircleCI or Taskcluster are supported for building and testing. We use Github Actions. Do we need to change CI?

jwhitlock:
Tracking the allows tools questions with PRs, as suggested by [CONTRIBUTE.md](https://github.com/mozilla-services/Dockerflow/blob/main/CONTRIBUTE.md):

[...]

Issue #88, opened 2021-03-10
Allow PRs from forked repositories to pass CI tests


jwhitlock:
PR #86 was from a fork of this repo, and failed on the ``build-image`` step with this log:

```
2021-03-10T01:29:56.2407941Z ##[group]Run aws-actions/configure-aws-credentials@v1
2021-03-10T01:29:56.2408709Z with:
2021-03-10T01:29:56.2409083Z   aws-region: us-west-2
2021-03-10T01:29:56.2409466Z env:
2021-03-10T01:29:56.2409814Z   AWS_REGION: us-west-2
2021-03-10T01:29:56.2410205Z ##[endgroup]
2021-03-10T01:29:56.8217681Z ##[error]Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers
```

The [configure-aws-credentials action](https://github.com/aws-actions/configure-aws-credentials ) is using [encrypted secrets](https://docs.github.com/en/actions/reference/encrypted-secrets) as recommended:

https://github.com/mozilla-it/ctms-api/blob/f3b561b0b8d825554e3c66dcd3a1aca9f8c6d3fd/.github/workflows/build-image.yaml#L39-L47

This may be a misconfiguration of our secrets, or we may be correctly blocking secrets for a forked repo PR. In either case, a forked repo PR should be able to pass CI.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-49)

jwhitlock:
A side effect of building the final image is running the linting steps. Linting should still run on pull requests from forked repositories.

grahamalama:
We no longer build the image in this repo, so I'll close this issue.

PR #90, opened 2021-03-10
Changing Dockerfile to listen on port 8000

duallain:
Will be worked on in more depth in https://github.com/mozilla-it/ctms-api/issues/89
but, our helm chart exports port 8000 at the moment, and since that's where we are planning on ending up
making this quick change now.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -95,4 +95,4 @@ ENTRYPOINT /docker-entrypoint.sh $0 $@
 
 EXPOSE 80
 
-CMD ["uvicorn", "ctms.app:app", "--host", "0.0.0.0", "--port", "80"]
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -95,4 +95,4 @@ ENTRYPOINT /docker-entrypoint.sh $0 $@
 
 EXPOSE 80
 
-CMD ["uvicorn", "ctms.app:app", "--host", "0.0.0.0", "--port", "80"]
+CMD ["uvicorn", "ctms.app:app", "--host", "0.0.0.0", "--port", "8000"]
```
bsieber-mozilla, line 98: lgtm

bsieber-mozilla (approved):
👍 

I see the fork issue has got ya here as well, I've adjusted perms and should be able to work in this repo until forking works with CI as per issue #88

PR #92, opened 2021-03-11
Move alembic to prod container

duallain:
To get migrations working in prod we need the prod container to have alembic
Rather than running a different container for migrating the db
moving the dependency to the prod container


jwhitlock:
There's a warning on build:

```
Installing dependencies from lock file
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.
```

Next time, run ``poetry update`` to get the lock file in sync with any manual changes, or use ``poetry add --dev alembic`` to have poetry update the dependencies.

Issue #51, opened 2021-02-24
Use selectinload, sort by name for newsletter query


jwhitlock:
Instead of manually making a second query, use the SQLAlchemy [selectinload](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html?highlight=selectinload#selectin-eager-loading) to fetch the related newsletter rows when getting a full contact.

Part of Jira:OSS-305

PR #94, opened 2021-03-12
Refactor complete contact queries

jwhitlock:
## Proposed changes

To fix issue #51:

* Refactor the main query from ``get_contact_by_email_id`` and ``get_contacts_by_any_id`` into ``_contact_base_query``, and use some SQLAlchemy [relationship loading techniques](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html) to load related data directly onto the ``Email`` object. This increases efficiency when multiple contacts are loaded for an alternate ID, changing from N+1 queries to 2, up to N=500 contacts.
* Return newsletters ordered by name, for consistent output.
* Add direct tests of CRUD methods, including a ``StatementWatcher`` for counting SQL queries and debugging issues.

Future work I talked myself out of for this round:

* Change ``get_contact_by_email_id`` to ``get_email_by_email_id``, and just return the ORM ``Email`` instance. This would change the application and some tests.
* Test all the CRUD methods and get 100% coverage.

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -1,7 +1,7 @@
 from typing import Dict, List, Optional
 
 from pydantic import UUID4, EmailStr
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, joinedload, selectinload
 
 from .auth import hash_password
 from .models import (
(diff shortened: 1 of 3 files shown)


imbstack (approved):
I love the testing for this

PR #95, opened 2021-03-12
Add POST /ctms to public API group

jwhitlock:
This groups it with the other public APIs in the interactive docs.


imbstack (approved):
thanks

PR #103, opened 2021-03-13
Adopt Dockerflow, other changes

### LINKED ISSUE #68: Test alembic migrations
Integrate [pytest-alembic](https://github.com/schireson/pytest-alembic) (ht @bsieber-mozilla) to test these scenarios:

* Are the models and migrations in sync?
* Is there a single alembic head? (can be broken if two database PRs were merged)
* (Future) Migrate up to a revision, add some problematic production-like data, run the remaining migrations

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-50)

jwhitlock:
## Proposed changes

For issue #89, adjust ``Dockerfile`` for [dockerflow](https://github.com/mozilla-services/dockerflow):

* Add an environment variable ``PORT``, default 8000, to control the port that the application is served from.
   - This required changing to ``ENTRYPOINT ["/docker-entrypoint.sh"]``, so that ``CMD`` could use shell expansion to get the value of ``$PORT``
* Add a user ``app``, default to user id 10001 and group id 10001, give it ownership of the application files, and run the application as the ``app`` user.
   - The ``app`` user can cause problems for Linux users, if it creates files owned by user 10001 that the development account can't read or delete. If this happens, there are ``.env`` overrides ``CTMS_UID`` and ``CTMS_GID`` to set these for local development.
* Add a generic ``version.json``, and populate it with build details when building the Docker image with a Github action.
* Add ``/__version__``, which returns the contents of ``version.json``
* Replace ``/health`` with ``/__lbheartbeat``, which returns that the application is up and running.
* Add ``/_heartbeat``, which checks backing services (currently just the database.
  - I think it should also return the alembic version, but I need to fix #68 first. 

Other changes not related to dockerflow:

* Update the Python image from 3.8.1 to 3.8.2
* Update poetry from 1.1.4 to 1.1.5
* Add ``PYTHONPATH=/app`` so that Python code can import from ``ctms``, and tools no longer require ``python -m`` to run.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -31,7 +31,7 @@ jobs:
           IMAGE_TAG: ${{ needs.prepare.outputs.FULL_IMAGE_TAG }}
         run: |
           echo "Building an image with the following tag:"
-          echo $IMAGE_TAG 
+          echo $IMAGE_TAG
 
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


jwhitlock:
I've added the ``/__version__``, ``/__heartbeat__``, and ``/__lbheartbeat__`` endpoints, so this PR now covers the required bits of Dockerflow, leaving MozLog for a later PR.

jwhitlock:
Rebased on ``main``, with variable name changes to satisfy ``pylint``

imbstack (approved):
makes sense to me!

bsieber-mozilla (approved):
Thanks for tackling this extensive change!

PR #109, opened 2021-03-16
Convert ``.first()`` to ``.one_or_none()`` or ``.all()

### LINKED ISSUE #50: Use .one_or_none when reading from DB by email_id
We're using ``.first()`` to read from the database, which applies a ``LIMIT 1`` to fetch just one record, and returns ``None`` if no rows are returned. [``.one_or_none()``](https://docs.sqlalchemy.org/en/13/orm/query.html?highlight=first#sqlalchemy.orm.query.Query.one_or_none) is a stronger version, also returning a row or ``None``, but raising a ``MultipleResultsFound`` exception if multiple rows are found.

Part of Jira:OSS-305

jwhitlock:
Fix #50 by converting ``get_amo_by_email_id``, ``get_fxa_by_email_id``, and ``get_vpn_by_email_id`` from ``.first()`` to ``.one_or_none()``, to detect early if there are multiple records for one ``email_id``.

Convert ``get_newsletters_by_email_id`` from ``.first()`` to ``.all()``, which mostly affected how to test if none were written.

Remove ``get_email_by_email_id``, which was unused and untested.

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -24,15 +24,19 @@ from .schemas import (
 
 
 def get_email_by_email_id(db: Session, email_id: UUID4):
-    return db.query(Email).filter(Email.email_id == email_id).first()
+    return db.query(Email).filter(Email.email_id == email_id).one_or_none()
 
 
 def get_amo_by_email_id(db: Session, email_id: UUID4):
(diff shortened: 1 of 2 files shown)


imbstack (approved):
thanks so much, I was dreading doing this one for some reason

cvalaas (approved):

PR #110, opened 2021-03-16
Allow testing outside of docker

imbstack:
## Proposed changes

The docker container is in UTC so the timestamps it generates are in
that timezone. The `datetime.utcnow()...timestamp()` in tests is kinda
funny. The note in
https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp
explains a bit more. Basically when I tried running pytest on my machine
(which is in PST currently), the result of that call was 7 hours in the
past instead of correctly in UTC.

Happy to be told this is wrong if I'm misundersanding something. I _think_ this
is right but timezones are hard!

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/tests/unit/test_auth.py
+++ b/tests/unit/test_auth.py
@@ -1,6 +1,6 @@
 """Test authentication"""
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 import pytest
 from jose import jwt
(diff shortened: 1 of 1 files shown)


bsieber-mozilla (approved):
Time related tests are no-fun. 

✅

imbstack:
Landing this now but please take a peek when you're back, @jwhitlock

jwhitlock:
Reading the Python docs, I shouldn't have used [utcnow](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow), which returns the current UTC date and time, but as a naive datetime object with no timezone. There's a big warning that naive date times are treated as local time, and recommends ``datetime.now(timezone.utc)`` instead. ``utcnow()`` seems dangerous and also a bit useless. I'll put together a PR to replace it.

Issue #105, opened 2021-03-15
Interactive Swagger does not apply credentials on 303 See Other from successful POST /ctms


jwhitlock:
When submitting a ``POST /ctms`` with this content:

```json
{
  "email": {
    "primary_email": "test-20210315@example.com",
    "email_id": "f6575c19-ae5b-428c-a2cc-74bb6ffd40c2"
  }
}
```

The Swagger UI returns a ``401 Unauthorized``:

```json
{
  "detail": "Could not validate credentials"
}
```

However, fetching the newly created record (``GET /ctms/f6575c19-ae5b-428c-a2cc-74bb6ffd40c2``) succeeds.

If it is done locally, you can see the ``303 See Other`` redirect from the server. It appears that the Swagger UI is following the redirect, but not adding the credentials to the new fetch, so it is impossible to tell if there was a credential issue or it was successful.

imbstack:
I guess it sorta makes sense to strip auth header before following a redirect.

jwhitlock:
Would you be OK with a 20x response, probably `201 Created`, with a `Location` header and the contact JSON as content?

bsieber-mozilla:
I think this can be closed, since the merge of pr #189

PR #112, opened 2021-03-19
Contact PUT

### LINKED ISSUE #55: Add PUT /ctms/{email_id} to insert or update a contact
Add a ``PUT /ctms/{email_id}`` endpoint that will insert or update a contact. The body is JSON that includes the contact data.

* The element ``email.primary_email`` ~~and ``email.basket_token``~~ is required, and an error (TBD) is returned if omitted. The ``email_id`` in the URL is a third ID.
  * If ``email.email_id`` is included in the data, it must match ``email_id`` in the URL, or an error (TBD) is returned
  * If an existing contact matches the ~~three~~ two IDs, it is considered an update.
  * If no existing contact matches any of the ~~three~~ two IDs, it is considered an insert. 
  * If an existing contact matches some but not all of the IDs, a ``409 Conflict`` error is returned.
* Other items are optional
* If the action is an insert:
  * If a group (like ``amo``) is sent with the default values, it is treated as if it was omitted (i.e., no related ``amo`` row is created)
  * The return is a ~~``201 Created``, and the body is the new contact JSON~~ ``303 See Other`` to ``/ctms/{email_id}`` (see #303)
* If the action is an update:
  * ~~If an item is omitted, the existing value for that item is unchanged~~ - Moved to ``PATCH``, #128 
  * If an item is included, the existing value is updated
  * If a group (like ``amo``) is sent with all default values, any existing row is deleted
  * ~~To delete a group, send TBD (``None``? ``{}``? ``{"delete": true}``?)~~ - Moved to ``PATCH``, #128 
  * The return is a ~~``200 OK``, and the body is the updated contact JSON~~ ``303 See Other`` to ``/ctms/{email_id}`` (see #303)

Tracked as part of Jira:OSS-305

imbstack:
## Proposed changes

Should approximately finish #55 (although there are a few discrepancies to clear up now that I look) Maybe worth doing in a followup though

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,7 @@ repos:
         name: pylint
         entry: pylint
         language: system
+        exclude: ^(migrations|tests)/
         types: [python]
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.1.0
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -21,6 +21,7 @@ from .auth import (
 )
 from .crud import (
     create_contact,
+    create_or_update_contact,
     get_api_client_by_id,
     get_contact_by_email_id,
     get_contacts_by_any_id,
@@ -32,6 +33,7 @@ from .schemas import (
     ApiClientSchema,
     BadRequestResponse,
     ContactInSchema,
+    ContactPutSchema,
     ContactSchema,
     CTMSResponse,
     EmailSchema,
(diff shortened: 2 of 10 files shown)


cvalaas (approved):

jwhitlock (changes requested):
Thanks @imbstack! I tried this in the swagger API. It is challenging due to issue #105, but my basic tests worked (add a minimal contact, update that contact).

The two big issues I see are the PUT path (``/ctms`` vs ``/ctms/{email_id}``), and the failing tests.

I envisioned this endpoint's strategy as "omitting a parameter means leave it alone", where instead it has the strategy of "make it look like what I sent you". That means that you always have to GET the existing object to update, and last writer wins. I think that is less useful for the way Basket uses it, but we can make it work.

jwhitlock (approved):
Thanks @imbstack!

PR #113, opened 2021-03-22
Convert utcnow() to now(utc)

jwhitlock:
[datetime.utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) returns the current UTC time as a naive datetime, which is often interpreted as a local time. The Python docs recommend using [datetime.now(timezone.utc)](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) instead. I'm having a hard time thinking of when ``utcnow()`` is appropriate.

This is follow-on work for PR #110. It works with ``make test``. I don't have my environment setup to test outside of docker.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Checklist

- N/A I have added tests that prove my fix is effective or that my feature works
- N/A I have added necessary documentation (if appropriate)

### DIFF
--- a/ctms/auth.py
+++ b/ctms/auth.py
@@ -7,7 +7,7 @@ header. A JWT token is returned that expires after a short time. To renew,
 the client POSTs to /token again.
 """
 import warnings
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Dict, Optional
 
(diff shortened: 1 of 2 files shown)


imbstack (approved):
ah, that makes much more sense. ty

PR #116, opened 2021-03-24
adding a puller-from-bigquery

cvalaas:
## Proposed changes

This is some code to pull SFDC data from BigQuery (Deltas or Fulls).
Some notes:
- Doesn't need to be merged in, if you want to incorporate it differently
- Always pulls from production data (could be changed)
- Requires a json file to authenticate with Google. I'm unsure how we're handling secrets at this point
- All the field names are in comments at the bottom of the file

### DIFF
--- a/ctms/bigquery-import.py
+++ b/ctms/bigquery-import.py
@@ -1,0 +1,105 @@
+from google.cloud import bigquery
+import os 
+
+if "GOOGLE_APPLICATION_CREDENTIALS" not in os.environ:
+    # TODO: put this in secrets somewhere
+    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = "mozilla-cdp-prod-810d22419b61.json"
+
+class BigQuerySFDCData():
+
+    def __init__(self):
+        self.bq_client = bigquery.Client()
+
+    def get_table_rows(self, table_name=None, delta=True):
+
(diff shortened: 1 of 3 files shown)


ctms/bigquery-import.py
```diff
@@ -0,0 +1,105 @@
...
+# double_opt_in          BOOLEAN NULLABLE 
+# has_opted_out_of_email BOOLEAN NULLABLE 
+# unsubscribe_reason     STRING  NULLABLE 
+# create_timestamp       STRING  NULLABLE 
+# update_timestamp       STRING  NULLABLE 
+#
+# CTMS_contact_to_amo
+# -------------------
+# email_id             STRING     NULLABLE     
+# amo_add_on_ids       STRING     NULLABLE
```
cvalaas, line 62: @imbstack I noticed that in the ctms/schemas/addons.py all these fields are defined without the "amo_" prefix (and the same goes for newsletters and fxa). Would it help if I changed the BQ views to similarly strip those prefixes so that we match the already-defined schemas?

imbstack: Ah, sorry I didn't see this when you posted it. I think either way is fine. If it isn't a lot of work to remove the prefix that is easiest but it should work either way?

imbstack:
Landed via #136

PR #117, opened 2021-03-24
Use scoped_session for database session

### LINKED ISSUE #70: Make sure we're doing sessions correctly
https://docs.sqlalchemy.org/en/13/orm/contextual.html

jwhitlock:
## Proposed changes

Follow [SQLAlchemy's advice](https://docs.sqlalchemy.org/en/13/orm/contextual.html#using-thread-local-scope-with-web-applications) to use [scoped_session](https://docs.sqlalchemy.org/en/13/orm/contextual.htm) to wrap the session factory. This may be unnecessary with FastAPI's dependency injection, but if we're wrong, it should prevent issues of multiple threads getting the same (not thread-safe) session.

Fixes #70

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist

- N/A I have added tests that prove my fix is effective or that my feature works
- N/A I have added necessary documentation (if appropriate)

## Further comments

We do not use the changed functions in tests, only in the web application, so there are no changes to tests. I ran the web application locally to check that the code worked without errors.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -9,7 +9,7 @@ from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
 from pydantic import EmailStr
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, scoped_session
 
 from . import config
 from .auth import (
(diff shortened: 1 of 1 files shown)


ctms/app.py
```diff
@@ -63,7 +63,8 @@ def get_settings():
 @app.on_event("startup")
 def startup_event():
     global SessionLocal  # pylint:disable = W0603
-    _, SessionLocal = get_db_engine(get_settings())
+    _, session_factory = get_db_engine(get_settings())
+    SessionLocal = scoped_session(session_factory)
```
bsieber-mozilla, line 67: Interesting!

bsieber-mozilla (approved):
LGTM

Issue #82, opened 2021-03-03
Move Mozilla Foundation data to a subgroup, split ID


jwhitlock:
Move the Mozilla Foundation data to a subgroup, to allow omitting it when not set, and to be more flexible about future additions and changes

* Add a ``mofo`` group, with:
  * Standard table stuff (ID, timestamps)
  * ``mofo_email_id`` (string 100, unique)
  * ``mofo_contact_id`` (string 100, can have dupes)
  * ``mofo_relevant`` (bool default false, legacy?)
* Remove fields from ``email`` group:
  * ``mofo_id``
  * ``mofo_relevant``
  
For now, these are CTMS-only fields, and won't be propagated to Acoustic.

Design tracked in Jira:OSS-316, implementation may be in future sprint.

Before the change, the JSON for a MoFo contact would look something like this:

```json
{
  "email": {
    "primary_email": "donor@example.com",
    "basket_token": "c4a7d759-bb52-457b-896b-90f1d3ef8433",
    "double_opt_in": true,
    "sfdc_id": "001A000023aABcDEFG",
    "mofo_id": "string",
    "first_name": "Jane",
    "last_name": "Doe",
    "mailing_country": "us",
    "email_format": "H",
    "email_lang": "en",
    "has_opted_out_of_email": true,
    "unsubscribe_reason": "string",
    "email_id": "332de237-cab7-4461-bcc3-48e68f42bd5c",
    "create_timestamp": "2020-03-28T15:41:00.000Z",
    "update_timestamp": "2021-01-28T21:26:57.511Z"
  },
  ...
}
```

and after:

```json
{
  "email": {
    "primary_email": "donor@example.com",
    "basket_token": "c4a7d759-bb52-457b-896b-90f1d3ef8433",
    "double_opt_in": true,
    "sfdc_id": "001A000023aABcDEFG",
    "first_name": "Jane",
    "last_name": "Doe",
    "mailing_country": "us",
    "email_format": "H",
    "email_lang": "en",
    "has_opted_out_of_email": false,
    "unsubscribe_reason": "string",
    "email_id": "332de237-cab7-4461-bcc3-48e68f42bd5c",
    "create_timestamp": "2020-03-28T15:41:00.000Z",
    "update_timestamp": "2021-01-28T21:26:57.511Z"
  },
  "mofo": {
    "mofo_email_id": "string",
    "mofo_contact_id": "other_string",
    "mofo_relevant": true,
  },
  ... 
}
```

[...]

PR #96, opened 2021-03-12
Make basket_token optional. Fixes #54

imbstack:
## Proposed changes

_Describe the big picture of your changes here..._

## Types of changes

What types of changes does your code introduce?
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

- [ ] I have read the [guides](https://github.com/mozilla-it/ctms-api/tree/main/guides)
- [ ] I have followed the [Mozilla Lean Data Policies](https://www.mozilla.org/en-US/about/policy/lean-data/)
- [ ] Lint and tests pass both locally and on the cicd with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

_Use this space for additional comments, complications, learnings, alternatives,..._

PR #118, opened 2021-03-26
Move Mozilla Foundation data to new mofo group

jwhitlock:
## Proposed changes

Move the Mozilla Foundation data to a new subgroup ``mofo``, to allow omitting the data when not set, and to be more flexible about future additions and changes. This fixes issue #82.

The group has these items:
* ``mofo_email_id``: MoFo's identifier for the email, unique, was ``email.mofo_id`` and non-unique
* ``mofo_contact_id``: MoFo's identifier for a contact, which may have several emails
* ``mofo_relevant``: This contact is relevant to MoFo, was ``email.mofo_relevant``

``mofo_email_id`` and ``mofo_contact_id`` are available as alternate IDs when calling ``GET /ctms``.

The database columns ``email.mofo_id`` and ``email.mofo_relevant`` will be removed after this code is in production (issue #119).

Because we don't have production data yet, the data migration is omitted.

These changes are mostly in the last commit. See below for info on the previous commits.

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- ✓ I have added tests that prove my fix is effective or that my feature works
- ✓ I have added necessary documentation (if appropriate)

## Further comments

I made several changes to prepare for this change, so that the final commit would be focused on the database, schema, and API changes:

* Add tests for ``ContactInSchema.idempotent_equal``, to improve coverage and add a place to test the new group.
* A straightforward addition of a new ``if`` clause resulted in a ``pylint`` warning, so I refactored the content of ``idempotent_equal`` as well. Python's [and](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not) is a short-circuit operator, so the new code should be just as efficient.
* Added the ``PUT`` API to the authentication tests, and adjusted the ``POST`` test to remove ``basket_token``, fixed in #96.
* Changed the ``GET /ctms`` API to return a list of ``CTMSResponse``, which requires all the subgroups, so that omitting ``mofo`` will be a pydantic error rather than verified by tests. The old ``CTMSResponse`` is now ``CTMSSingleResponse``.
* Split up the expected error message for omitting parameters to ``GET /ctms``, so that ID changes will be easier to read in diffs
* My editor runs ``pylint`` on all files, including ones omitted by project tests. I saw a few warnings in unit test files that looked legit, and made the small changes
* Added ``--autogenerate`` to the ``alembic`` command to generate migrations
* Adjusted the migrations template file so that it doesn't trigger ``pylint``, ``detect-secrets``, or ``isort``. It still needs a ``black`` pass for indenting, and I like to remove ``alembic``s comments about changing the output.


ctms/models.py
```diff
@@ -24,13 +24,13 @@ class Email(Base):
     primary_email = Column(String(255), unique=True, nullable=False)
     basket_token = Column(String(255), unique=True)
     sfdc_id = Column(String(255))
-    mofo_id = Column(String(255))
+    mofo_id = Column(String(255))  # TODO: Remove
```
imbstack, line 27: can we add the issue id for removing this here?

imbstack (approved):
ty looks great

guides/developer_setup.md
```diff
@@ -195,20 +195,14 @@ To create a new migration file based on updated SQLAlchemy models:
 
 ```sh
 make shell  # To enter the web container
-alembic revision -m "A short description of the change"
+alembic revision --autogenerate -m "A short description of the change"
+black /app/migrations/versions/
```
bsieber-mozilla, line 199: Thanks for the updates here

bsieber-mozilla (approved):
lgtm, thanks!

PR #121, opened 2021-03-29
Working to get tagging correct for all builds

duallain:
## Proposed changes

Changing tagging to -

be {shortsha} on all commits that don't match the other rules
be "stg-{shortsha}" for commits merged to main.
be "v{semver}" (ex v.0.0.1) for production deploys. (this should work for releases cut and annotated tags pushed via command line)

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [guides](https://github.com/mozilla-it/ctms-api/tree/main/guides)
- [x] I have followed the [Mozilla Lean Data Policies](https://www.mozilla.org/en-US/about/policy/lean-data/)
- [x] Lint and tests pass both locally and on the cicd with my changes
- [x] I have added necessary documentation (if appropriate)

## Further comments

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -15,7 +15,7 @@ jobs:
           echo $CI_COMMIT_SHORT_SHA;
           echo ${GITHUB_REF##*/};
           echo $version_pattern;
-          if [[ ${GITHUB_REF##*/} =~ $version_pattern ]]; then
+          if [[ ${GITHUB_REF} =~ $version_pattern ]]; then
             echo "::set-output name=tag::${GITHUB_REF##*/}"
           elif [ ${GITHUB_REF##*/} = "main" ]; then
             echo "::set-output name=tag::stg-$CI_COMMIT_SHORT_SHA"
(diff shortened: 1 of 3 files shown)


.github/workflows/build-image.yaml
```diff
@@ -6,19 +6,25 @@ jobs:
...
+      - name: Set Tag
+        id: tag
         run: |
-          if [ ${GITHUB_REF##*/} = "main" ]; then
-            echo "::set-output name=prefix::stg-"
+          export CI_COMMIT_SHORT_SHA=$(echo $GITHUB_SHA | cut -c1-8)
+          echo $CI_COMMIT_SHORT_SHA;
+          echo ${GITHUB_REF##*/};
+          echo $version_pattern;
+          if [[ ${GITHUB_REF##*/} =~ $version_pattern ]]; then
```
jwhitlock, line 18: I think this will also push branches named ``v0.0.0`` to production.

* Branch `refs/heads/v0.0.0`
* Tag `refs/tags/v0.0.0`
* PR `refs/pull/121/merge`

I think we also need to match on the `/tags/` part, not just the last component.

jwhitlock (changes requested):
I think some small changes are needed to only push tags to production, and avoid pushing cleverly named branches.

duallain:
@jwhitlock updated with your suggestion to match on tags - I went with a minimal match of just tags/ rather than the full string, but hopefully does what you're suggesting.

PR #130, opened 2021-04-07
Test and improve ingestion

imbstack:
## Proposed changes

This should cover (maybe not at a perfect level) much of the ingestion code in tests. It also changes how the ingestion uses the db such that all of the ingestion happens with sqlalchemy core and inside a single transaction. This speeds things up significantly

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/ctms/bin/load_sample_csv.py
+++ b/ctms/bin/load_sample_csv.py
@@ -10,7 +10,7 @@ from typing import Any, Callable, Dict
 from uuid import uuid4
 
 from pydantic import BaseModel
-from sqlalchemy.orm import Session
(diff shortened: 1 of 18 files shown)


cvalaas (approved):

jwhitlock (approved):
Thanks @imbstack, the code and tests look good, and it worked for me in the local dev environment.

I had a non-migrated database, so I got a very large error message about `sqlalchemy.exc.DataError: (psycopg2.errors.StringDataRightTruncation) value too long for type character varying(10)`, followed by kilobytes of SQL. I think we only plan to use this during the initial migrations, not in the long term, so I think the error is acceptable.

Issue #128, opened 2021-04-06
Add PATCH /ctms/{email_id} to partially update a contact


jwhitlock:
Add a ``PATCH /ctms/{email_id}`` endpoint that will partially update a contact. The body is JSON that includes the contact data to update.

* If an item is omitted, the existing value for that item is unchanged
* If an item is included, the existing value is updated, unless it would create an ID conflict
* If a group (like ``amo``) is sent with all default values, any existing row is deleted
* To delete a group, send TBD (``None``? ``{}``? ``{"delete": true}``?)
* The return is a ``200 OK``, and the body is the complete updated contact JSON

Tracked in Jira:OSS-383

jwhitlock:
Implemented in PR #131 and PR #133:

* ID conflicts return a ``409 Conflict``
* Unsubscribe is ``{"newsletters": "UNSUBSCRIBE"}``
* Deletion is ``{"amo": "DELETE"}``
* Return is a 303. I'm planning on fixing all of them at once (#105)

PR #131, opened 2021-04-08
Add PATCH /ctms/{email_id}

jwhitlock:
As part of issue #128, add a ``PATCH /ctms/{email_id}`` API.

* The request is similar to ``PUT /ctms/{email_id}``, but the object must already exist
* It can't be used to change the ``email_id``
* Only the included items are updated; excluded items keep the existing values
* If a group is set to all default values, it is deleted from the database
* Newsletter subscription can be done by adding it to the ``newsletters`` list. It is not required to send all the user's subscriptions, but Basket probably will anyway.
* Newsletter unsubscription is done by setting ``"subscribed": False`` for that subscription, ``name`` is required. 

Differences from issue #128:

* I used a ``303 See Other`` redirect instead of a ``200``. I think a 2xx response is still a good idea to work better with the interactive docs repo, but I think some further changes are needed to make that easy, so it should be a new PR.
* I didn't handle cases of unique ID conflicts, it currently is a 5xx error
* I didn't add a way to delete a group, other than setting everything to defaults. Basket may need this to support some private APIs
* I didn't add a way to unsubscribe from all newsletters. This could be hacked on the Basket side, but something similar to group deletion may be more elegant.

I'm planning to address these in future PRs, but I'm open to including any of these changes in this PR.

- ✓ New feature (non-breaking change which adds functionality)

## Checklist

- ✓ I have added tests that prove my fix is effective or that my feature works
- *N/A* I have added necessary documentation (if appropriate)
- *N/A* Any dependent changes have been merged and published in downstream modules

## Further comments

``pylint`` runs on my test files, despite them being skipped in ``scripts/lint.sh``. Some changes were suggested by linting, like ``watcher`` instead of ``sw`` for ``StatementWatcher``, and keeping ``test_api.py`` from exceeding 1000 lines by adding the tests to a new file ``test_api_patch.py``.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -24,9 +24,11 @@ from .crud import (
     create_or_update_contact,
     get_api_client_by_id,
     get_contact_by_email_id,
-    get_contacts_by_any_id,
+    get_email,
+    get_emails_by_any_id,
 )
 from .database import get_db_engine
+from .models import Email
 from .monitor import check_database, get_version
 from .schemas import (
     AddOnsSchema,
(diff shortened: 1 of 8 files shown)


imbstack (approved):
makes sense to me. I think if I have a few minutes soon I'm going to disable pylint's "this is a bad name" thing or switch to flake8

jwhitlock:
> makes sense to me. I think if I have a few minutes soon I'm going to disable pylint's "this is a bad name" thing or switch to flake8

I have a fairly low confidence in my own naming conventions, I'm happy to follow the advice of any robot for consistency. If you have stronger opinions, let's switch or tweak settings.

Issue #132, opened 2021-04-09
Support refresh token workflow


jwhitlock:
Basket uses [requests-oauthlib](https://requests-oauthlib.readthedocs.io/en/latest/), which, as expected, integrates ``requests`` with ``oauthlib``. For common authentication implementations, it supports [automatic refreshing of access tokens](https://requests-oauthlib.readthedocs.io/en/latest/oauth2_workflow.html#refreshing-tokens). This requires a URL for requesting refresh tokens that accepts ``grant_type=refresh_token``. In many flows, a refresh token is used to request the new access token, but a ``client_id`` and ``client_secret`` can be used as well, and would be compatible with our ``client_credentials`` workflow.

Implementing token refresh would let the Basket code rely more on requests-oauthlib for automatic refreshing of tokens, simplifying the Basket-specific code.

One way to implement this would be for ``/token`` to accept ``grant_type=refresh_token``, in addition to ``grant_type=client_credentials``, and treat it the same as an initial request for an access token.

A second way would be to have a new endpoint ``/token/refresh`` (or similar) to handle this case.

requests-oauthlib is flexible enough to refresh without a refresh token (I think), so my preference is to not generate one.

jwhitlock:
I tested the ``requests-oauthlib`` refresh token workflow in the basket code in dev, looks good.

PR #134, opened 2021-04-09
Add grant=refresh-token to /token

jwhitlock:
Allow ``grant=refresh-token`` as well as ``grant=client-credentials`` for the ``/token`` endpoint to get an access token. The ``client_id`` and ``client_password`` are still required, and a ``refresh_token`` is not provided, required or checked. This gives it the same security as granting the original token.

This fixes issue #132, which explains why Basket will use this code.


bsieber-mozilla (approved):
LGTM

Issue #140, opened 2021-04-16
Add `after` to `GET /updates` response JSON, allow empty


jwhitlock:
The user of the `GET /updates` API has two requested changes:

1. Add the `after` parameter to the JSON response, so that the next URL can be constructed from the JSON:

```json
{
  "next": "https://ctms.example.com/updates?start=2021-03-31T20:30:00+00:00&end=2021-03-31T20:45:00+00:00&after=MjAyMS0wMy0zMVQyMDozMToyOC40MzE0MDIrMD&limit=10",
  "start": "2021-03-31T20:30:00+00:00",
  "end": "2021-03-31T20:45:00+00:00",
  "limit": 10,
  "after": "MjAyMS0wMy0zMVQyMDozMToyOC40MzE0MDIrMD",
  "items": [
    {
      ... // CTMS contact JSON
    }, ... // Additional CTMS contact JSON
  ]
}
```

(this was in @bsieber-mozilla's initial PR, I asked him to remove it, so +1 points to beiber)

2. Allow an empty `after` parameter, such as:

https://ctms.example.com/updates?start=2021-03-31T00:00&after=

I'm guessing that all empty parameters, except for `start` should also work:

https://ctms.example.com/updates?start=2021-03-31T00:00&end=&limit=after=

bsieber-mozilla:
I think with the merging of #143 we can close this.

PR #143, opened 2021-04-19
Updating api to allow blank strs for query params, updating response to include 'after', updating tests

bsieber-mozilla:
#140 ref

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -239,23 +239,23 @@ def get_bulk_contacts_by_timestamp(
 
     if last_page:
         # No results/end
-        url_safe_encoded_str = None
+        after_encoded = None
         next_url = None
     else:
         last_result: CTMSResponse = results[-1]
-        url_safe_encoded_str = compressor_for_bulk_encoded_details(last_result)
+        after_encoded = compressor_for_bulk_encoded_details(last_result)
         next_url = (
             f"{get_settings().server_prefix}/updates?"
             f"start={start_time.isoformat()}"
             f"&end={end_time.isoformat()}"
             f"&limit={limit}"
-            f"&after={url_safe_encoded_str} "
+            f"&after={after_encoded} "
         )
 
     return CTMSBulkResponse(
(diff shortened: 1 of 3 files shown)


ctms/app.py
```diff
@@ -255,12 +255,20 @@ def get_bulk_contacts_by_timestamp(
...
         end=end_time,
+        after=url_safe_encoded_str,
         limit=limit,
         items=results,
         next=next_url,
     )
 
 
+def updates_helper(value, default):
+    blank_vals = ["", " ", None]
```
jwhitlock, line 266: It seems to me that `blank_vals = ["", None]` should be the list. Was a space (the string `" "`) needed to make this work? If it was, then should any whitespace string be permitted?

jwhitlock (approved):
Thanks @bsieber-mozilla, this worked well in my local testing.

ctms/app.py
bsieber-mozilla: I can remove this! I wasn't sure if there were any limitations on the cinchy side; a single whitespace would be allowed when this is removed, but would likely break the system. So this was a safety net.

ctms/app.py
```diff
@@ -255,12 +255,20 @@ def get_bulk_contacts_by_timestamp(
     return CTMSBulkResponse(
         start=start_time,
         end=end_time,
+        after=url_safe_encoded_str,
```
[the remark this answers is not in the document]
bsieber-mozilla: Sure, I didn't name this variable in this PR, just reused it, but I will do the needful!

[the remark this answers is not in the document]
bsieber-mozilla: `after_encoded` is the adjusted name

ctms/app.py
bsieber-mozilla: Removed the whitespace to just have blank_vals be `["", None]`

jwhitlock (approved):
Thanks for the nit fixes!

Issue #139, opened 2021-04-14
Add Sentry exception capture


jwhitlock:
Add a middleware to catch exceptions and report to [Sentry](https://sentry.prod.mozaws.net). Work with ops to add Sentry buckets for stage and production, and to add the team to them.

Part of Jira:OSS-333

jwhitlock:
The sentry integration has been deployed. It was interfering with tests, which required moving initialization into the FastAPI setup (which doesn't run in tests), in PR #150.

@duallain added the Sentry configuration to environment settings, so everything should work. We didn't get any natural exceptions, so I forced the issue by getting a shell in staging, and running inside Python:

```python
from sentry_sdk import init, capture_message
init()
capture_message("I am a message from staging")
```

This message showed up in the Sentry project, so I'm more confident that exceptions will as well.

However, we got a handful of 500 errors from sentry stage to basket dev, without a Sentry traceback for CTMS. I'm digging in  to see if I can determine the issue.

jwhitlock:
Fixed, deployed to stage and prod

PR #145, opened 2021-04-20
Add Sentry middleware, rely on server-side filtering

jwhitlock:
For issue #139:

* Add ``sentry-sdk``
* Initialize it from ``SENTRY_SDK`` and with the commit number from ``version.json``.
* Wrap the FastAPI app with the [ASGI middleware](https://docs.sentry.io/platforms/python/guides/asgi/). This is done in a new variable because the Sentry middleware doesn't expose the wrapped app's interface, and to make ``mypy`` happier.
* Use the wrapped Sentry app in dev and production

I've added a new setting ``CTMS_SENTRY_DEBUG`` to turn on debug-level logging for Sentry, which makes for a noisy initialization and shutdown.

To test it locally, I added these settings in `.env`:

```sh
CTMS_SENTRY_DEBUG=1
SENTRY_DSN=https://bad-9c7b6b6bcb7a45d7ae01fe8f60594083@sentry.127.0.0.1.nip.io/0
```

and then ran ``make test`` and ``make start``, and adding a line like ``raise Exception("Testing Sentry")`` to a view (such as ``version()``. The ``SENTRY_DSN`` enables Sentry, but it is a fake connection string (to local host) so it won't be able to send any data, but prints a lot of details about what it is trying to do. The deployment environments will have the real ``SENTRY_DSN`` values.

A few years ago, I implemented client-side filtering in https://github.com/mozilla-services/socorro/pull/4917. That was a lot of work, and would look similar to the monitoring code. I'd like to try the server-side filtering for CTMS. This means that some security details like authorization headers may be sent, but the Sentry server will obfuscate them. I feel better about this than logging because Mozilla is running our own Sentry server. But, I'm open to implementing client-side filtering instead.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -6,11 +6,13 @@ from typing import Dict, List, Optional, Tuple, Union
 from uuid import UUID, uuid4
 
 import dateutil.parser
+import sentry_sdk
 import uvicorn
 from fastapi import Depends, FastAPI, HTTPException, Path, Request, Response
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
 from pydantic import EmailStr
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session, scoped_session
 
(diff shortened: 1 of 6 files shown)


bsieber-mozilla (approved):
Had to run a `make build` before the `make {test,start}`. Saw the Sentry logs in `test` as well as `start`

LGTM!

PR #147, opened 2021-04-21
Imports actually work now

imbstack:
Each of these changes is relatively self-contained in a commit.

- Allow configuring bq tables
- Stop modifying prod data as if it is test data
- Initial import followups
- Allow loading csvs in prod
- Give up on validating emails for now
- Make debugging imports a bit easier
- Handle duplicates in ingestion
- Move csv modifiers into helper
- Add scripts for modifying csvs

In particular, I'd say checking that the removal of EmailStr makes sense. Much of this is just dumping in semi-horrible scripts that I used to make the imports work. This is just for later use if we need them. I make no promises about them not being TODO-riddled junk.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -6,15 +6,14 @@ from typing import Dict, List, Optional, Tuple, Union
 from uuid import UUID, uuid4
 
 import dateutil.parser
+import sentry_sdk
 import uvicorn
 from fastapi import Depends, FastAPI, HTTPException, Path, Request, Response
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session, scoped_session
-
-import sentry_sdk
-from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 
 from . import config
 from .auth import (
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -6,15 +6,15 @@ from typing import Dict, List, Optional, Tuple, Union
 from uuid import UUID, uuid4
 
 import dateutil.parser
-import sentry_sdk
 import uvicorn
 from fastapi import Depends, FastAPI, HTTPException, Path, Request, Response
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
-from pydantic import EmailStr
-from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session, scoped_session
+
+import sentry_sdk
+from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
 
 from . import config
 from .auth import (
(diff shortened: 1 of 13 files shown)


cvalaas (approved):

ctms/app.py
```diff
@@ -124,14 +124,14 @@ def get_contact_or_404(db: Session, email_id) -> ContactSchema:
 
 def all_ids(
     email_id: Optional[UUID] = None,
-    primary_email: Optional[EmailStr] = None,
+    primary_email: Optional[str] = None,
```
jwhitlock, line 127: I think we talked about this in other channels, but not in repo.

There are emails that fail pydantic / python-email-validator, such as `foo.@bar.com`. but are in the Salesforce / BigQuery list. I think the validator is correct, and these are invalid emails, potentially undeliverable, and including them in the our systems reduces the usefulness of the data.

There are also records with no emails, which seems even less useful than a invalid, perhaps mistyped email.

I think the right thing is to relax the constraints, like we're doing here, to get over the hump of the SFDC → Acoustic + CTMS transition. We also need an issue for the follow-on work of classifying the suspicious emails and blank values, and tightening up the acceptance of new data.

My hope was that we could be strict with new data and loose with existing data, but that is non-trivial work that has to be sacrificed to the deadline, and this is the right approach.

ctms/bin/prepare_csvs.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
+"""Given a dump of csvs from bigquery, prepare it for being imported
+into postgres with COPY."""
```
jwhitlock, line 3: Nit: I was surprised that the linter was OK with this style. I'd expect:
```suggestion
"""
Given a dump of csvs from bigquery, prepare it for being imported
into postgres with COPY.
"""
```

ctms/bin/report_duplicates.py
```diff
@@ -0,0 +1,66 @@
...
+        for f in os.listdir(directory):
+            if "contact_to_email" in f:
+                path = os.path.join(directory, f)
+                with open(path, "r", newline="") as csvfile:
+                    reader = csv.DictReader(csvfile)
+                    for i, line in enumerate(reader):
+                        print(line["primary_email"], line["email_id"], file=col)
+                        total += 1
+                        if i % 100000 == 0:
+                            print(total, 87339348, int((total / 87339348.0) * 100))
```
jwhitlock, line 40: Nit: move `87339348` into a variable

tests/unit/test_bin_load_sample_csv.py
```diff
@@ -19,7 +19,17 @@ def test_data_dir(request):
...
-    ret = main(connection, settings, ["-d", test_data_dir])
+    ret = main(
+        connection,
+        settings,
+        [
+            "-d",
+            test_data_dir,
+            "--dev",
+            "--duplicates",
+            f"{test_data_dir}/duplicates.txt",
```
jwhitlock, line 30: Nit: Can we put some data in the test duplicates, to get some coverage on the duplicate processor, even if the duplicate data isn't used in tests 

*I can't comment on the empty files, so I'm putting it here*

jwhitlock (approved):
Thanks @imbstack, this is a ton of work! I can see some follow-on work (converting print to logging, future email validation, etc.) but I think we should merge and deal with that later.

There's a couple of failing tests, and black wants to make some changes, and linting is broken for me. I'm seeing issues that are probably related to my recent commits adding sentry and test linting. I'm not sure the best way to proceed - collaboration on figuring out the problem, or merge and then let me fix it. I'm approving so you can pick the second tactic.

The test failure is from processing the newsletter loop. The input is a list of dictionaries, and ingest fails to call ``.dict()`` on a dictionary.

jwhitlock:
I dug a little into my `pylint` issues. They go away if I remove `sentry_sdk.init` from `ctms/app.py`. They appear in branch for #145 and in the `main` branch. I'm surprised that linting doesn't run in CI.

I think the quickest fix is to revert the changes to `pyproject.yml` and `scripts/lint.sh` from https://github.com/mozilla-it/ctms-api/commit/e7e7d708f0ea55f4cecd2c2cf51a56c6d5bddf3a, so that tests are no longer linted. I think the code changes in the tests to fix linting are OK. I could do this in a separate PR, before or after this merges, or it could be a commit in this PR.

For the long-term fix, I'll need to re-arrange how ``sentry_sdk.init`` is called. It may be enough to move it into ``startup_event``. It may be necessary to call it from any scripts as well.

tests/unit/test_bin_load_sample_csv.py
imbstack: Tracked in #148

ctms/app.py
imbstack: Yeah, agreed. I think I feel much better now that this data is in the db successfully. We can start processing it shortly.

PR #156, opened 2021-04-26
Updating timestamps for db updates

bsieber-mozilla:
### Some major modifications:
 Patch and Put will now update the timestamps for both the email/contact record, as well as the table being updated. 

This would allow the bulk read to function as is, still using the main contact/email record's timestamp as the indicator for any updates for a contact record or associated table.


ctms/schemas/addons.py
```diff
@@ -83,6 +83,14 @@ class Config:
 AddOnsInSchema = AddOnsBase
 
 
+class UpdatedAddOnsInSchema(AddOnsInSchema):
```
bsieber-mozilla, line 86: Added new instead of into existing objects because I didnt want the timestamps as an option for api calls. Allowing this hidden model as a segway between api input and db input.

bsieber-mozilla: `default_factory` is required over `default` for dynamic default values

jwhitlock (approved):
Thanks @bsieber-mozilla! The code and tests look correct. I tried it out locally, with ``PUT`` and ``PATCH``, and it appears to do the right thing.

jwhitlock (approved):
Thanks for the index on the ``update_timestamp``!

PR #157, opened 2021-04-26
Store import script final state

imbstack:
Just for posterity's sake. Hopefully we won't need these soon.


bsieber-mozilla (approved):
lgtm

PR #161, opened 2021-04-28
Updating indexes for bulk read operation

bsieber-mozilla:
-Removal of single column index: Email.update_timestamp
+Addition of composite index: Email.update_timestamp, Email.email_id

_Includes an autogenerated migration script._


jwhitlock (approved):
Thanks @bsieber-mozilla!

PR #166, opened 2021-04-28
Revert "Use scoped_session for database session"

jwhitlock:
This reverts commit 60985aedffb8827d6057c8fadb5772401d568008 from PR #117, to address re-opened issue #70.

FastAPI shares threads between requests, which causes issues for ``scoped_session``'s per-thread sessions.

@imbstack and @bsieber-mozilla have been able to reproduce in ctms-stage by making many similar requests in parallel.


imbstack (approved):
I think that makes sense?

PR #175, opened 2021-04-30
Updating bulk query for mofo_relevant

bsieber-mozilla:


cvalaas (approved):

ctms/crud.py
```diff
@@ -79,25 +79,45 @@ def _contact_base_query(db):
     )
 
 
+def query_filter_list(start_time, end_time, after_email_uuid, mofo_relevant):
```
jwhitlock, line 82: Nit: I'd do this work ``get_bulk_contents``, since it doesn't seem like a generally useful function.

ctms/crud.py
```diff
@@ -79,25 +79,45 @@ def _contact_base_query(db):
...
         after_email_uuid = uuid.UUID(after_email_id)
+    filter_list = query_filter_list(
+        start_time=start_time,
+        end_time=end_time,
+        after_email_uuid=after_email_uuid,
+        mofo_relevant=mofo_relevant,
+    )
+    bulk_contacts = _contact_base_query(db)
+    for query_filter in filter_list:
+        bulk_contacts = bulk_contacts.filter(query_filter)
```
jwhitlock, line 117: The tests confirm that this is applied as an ``AND`` filter, nice work

jwhitlock (approved):
@bsieber-mozilla thanks for the quick work! I'd like to see a test for an update with the mofo_relevant flag set, but it seems clear it will work on inspection.

ctms/crud.py
bsieber-mozilla: renamed this

ctms/crud.py
```diff
@@ -79,25 +79,45 @@ def _contact_base_query(db):
...
+        return [
+            Email.update_timestamp >= start_time,
+            Email.update_timestamp < end_time,
+            Email.email_id != after_email_uuid,
+        ]
+    return [
+        Email.update_timestamp >= start_time,
+        Email.update_timestamp < end_time,
+        Email.email_id != after_email_uuid,
+        Email.mofo.has(mofo_relevant=mofo_relevant),
```
bsieber-mozilla, line 93: `.has` was essential for truly getting the query to be properly created (printed sql query to visually confirm)

tests/unit/test_crud.py
```diff
@@ -186,6 +186,64 @@ def test_get_contact_by_email_id_miss(dbsession):
...
+    with StatementWatcher(dbsession.connection()) as watcher:
+        bulk_contact_list = get_bulk_contacts(
+            dbsession,
+            start_time=after_start,
+            end_time=end_time,
+            limit=10,
+            mofo_relevant=mofo_relevant_flag,
+        )
+    assert watcher.count == 1
+    assert len(bulk_contact_list) == 0
```
jwhitlock, line 214: This surprised me, I was expecting two results. I think I understand why - there's no row in the mofo table, so the query skips those, even though if you fetch those contacts it will say ``mofo_relevant = false``.

I think either the code should also include contacts with no mofo row, or we should forbid setting the value to false, so that it can only be used to select mofo relevant records and not irrelevant ones

bsieber-mozilla:
This PR introduces a new query param `mofo_relevant` that is optional, if provided as True or False, it will be provided to the underlying query. 

For MOFO relevant records, send mofo_relevant=True in the query params
For non mofo relevant records, send mofo_relevant=False
for all records, send nothing in mofo_relevant

tests/unit/test_crud.py
cvalaas: I agree, I think?
`mofo_relevant=True` should return all records with that flag set to true in the mofo table, while `mofo_relevant=False` should return all records where (that flag is false OR flag doesn't exist)

bsieber-mozilla: `mofo_relevant=None` imo should also be a possibility, but without the added filtration.

jwhitlock: Does ``mofo_relevant=None`` get set by ``/updates?mofo_relevant=`` (no value)? If so, then yes, that should act the same as if you called ``/updates`` without the new parameter.

bsieber-mozilla: no value... I think we'll need to make a minor shift to the api to allow for "" blank string instead of bools again. But yeah let me make that change

ctms/crud.py
```diff
@@ -79,25 +79,55 @@ def _contact_base_query(db):
     )
 
 
+def get_bulk_query(start_time, end_time, after_email_uuid, mofo_relevant):
```
jwhitlock, line 82: Now that there are three paths, I think there is a higher risk of common changes not being reflected across all three options. Can you combine the common path, something like this:

```python
filters = [
  Email.update_timestamp >= start_time,
  Email.update_timestamp < end_time,
  Email.email_id != after_email_uuid
]
if mofo_relevant is False:
  filters.append(_or(Email.mofo == None, Email.mofo.has(mofo_relevant=mofo_relevant))
if mofo_relevant is True:
  filters.append(Email.mofo.has(mofo_relevant=mofo_relevant))
return filters
```

pyproject.toml
```diff
@@ -76,6 +76,7 @@ testpaths = [
         "C0114", # mozilla-it/ctms-api#100
         "C0115", # mozilla-it/ctms-api#100
         "C0116", # mozilla-it/ctms-api#100
+        "C0121", # Because it's actually not a singleton, but a model comparison
```
jwhitlock, line 79: I'm not sure which like is giving this warning, but could just that line disable the warning?  The line syntax is ``# pylint: disable foo-message`` or ``# pylint: disable C0121``

bsieber-mozilla: I've moved this inline; but it was being called out near the filtration query for the `==` when it believes singletons are being compared

jwhitlock (approved):
Thanks @bsieber-mozilla, I think it is ready to merge!

bsieber-mozilla:
merging!

Issue #177, opened 2021-05-03
Enable more mypy


jwhitlock:
As an ongoing task, enable mypy in stricter modes to expand static checking. I should be possible for 100% of the CTMS code to be statically checked, but that some dependencies will not be checked.

mypy uses ``mypy.ini`` for options, and allows per-folder and per-file overrides in this file. mypy does not (yet) support ``pyproject.toml``

PR #178, opened 2021-05-03
Add mypy.ini, enabled warn-return-any

jwhitlock:
As part of increasing mypy strictness (issue #177), switch to ``mypy.ini``, which makes it easier to override settings for particular files, and to share the configuration between the pre-commit hook and the linting script.

Add the ``warn-return-any`` option, which warns when a type with ``Any`` is returned from a function that declares a non-``Any`` return type. The default for non-annotated functions is to return ``Any``, and this includes libraries where mypy can't see the annotations. For ``ctms/crud.py``, this exposed some issues with return types for functions that can return ``None``. For ``ctms/auth.py``, it mostly involved specifying the types for return values from library functions.

### DIFF
--- a/ctms/auth.py
+++ b/ctms/auth.py
@@ -48,7 +48,7 @@ def create_access_token(
     to_encode = data.copy()
     expire = (now or datetime.now(timezone.utc)) + expires_delta
     to_encode["exp"] = expire
-    encoded_jwt = jwt.encode(to_encode, secret_key, algorithm="HS256")
+    encoded_jwt: str = jwt.encode(to_encode, secret_key, algorithm="HS256")
     return encoded_jwt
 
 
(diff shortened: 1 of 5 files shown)


ctms/auth.py
```diff
@@ -133,7 +133,12 @@ def __init__(
 
     async def __call__(self, request: Request) -> Optional[str]:
         authorization: str = request.headers.get("Authorization")
-        scheme, param = get_authorization_scheme_param(authorization)
+
+        # TODO: Try combining these lines after FastAPI 0.61.2 / mypy update
+        scheme_param = get_authorization_scheme_param(authorization)
+        scheme: str = scheme_param[0]
+        param: str = scheme_param[1]
```
jwhitlock, line 140: I didn't understand why this was necessary. FastAPI defines [get_authorization_scheme_param](https://github.com/tiangolo/fastapi/blob/22528373bba6a654323de416ad5c867cbadb81bb/fastapi/security/utils.py#L4-L8) as returning ``Tuple[str, str]``, but mypy doesn't recognize the two-item assignment as ``str``. It could be a FastAPI packaging issue (a ``py.typed`` file may be missing), or a mypy issue (FastAPI pins the mypy version). I left a note to see if the one-line version will be valid in the future.

bsieber-mozilla (approved):
lgtm 🚀

PR #181, opened 2021-05-03
READY: Acoustic pending updates - Basic Operations + Scheduling of PendingRecords

bsieber-mozilla:
## Edit
New updates.

PR Includes, tests, and updates from the prior review. 

Records also should be scheduled on create/updates events.

----
Feature Branch - Pending Updates.

Still in testing. Throwing this out for a preliminary look.

- Does the table columns look at expected by others? Should we add any other fields?
- Do the CRUD operations seem useful? Necessary?
- Any other NITs or thoughts?


bsieber-mozilla:
Still in draft; would be helpful to get an early pass by reviewers and other acoustic parties to determine proper path and ability to split potentially.

bsieber-mozilla:
Next Steps:

- [ ] Test the CRUD operations (this PR)
- [ ] Input the records into the table on contact put/patch/post (next PR)

ctms/crud.py
```diff
@@ -267,6 +268,86 @@ def get_contacts_by_any_id(
...
+    db: Session,
+    end_time: datetime,
+):
+    """Get all the pending records up to a certain date."""
+    pending_records: List[PendingAcousticRecord] = (
+        db.query(PendingAcousticRecord)
+        .filter(
+            PendingAcousticRecord.update_timestamp < end_time,
+        )
+        .order_by(asc(Email.update_timestamp), asc(Email.email_id))
```
jwhitlock, line 281: Is this supposed to be ordered by ``PendingAcousticRecord.update_timestamp``, ``PendingAcousticRecord.email_id``?

ctms/crud.py
```diff
@@ -267,6 +268,86 @@ def get_contacts_by_any_id(
...
+        email_id=email_id, **updated_pending_record.dict()
+    )
+    stmt = stmt.on_conflict_do_update(
+        index_elements=[PendingAcousticRecord.email_id],
+        set_=updated_pending_record.dict(),
+    )
+    db.execute(stmt)
+
+
+def delete_pending_record(db: Session, email_id: UUID4):
```
cvalaas, line 335: This looks like it would delete all "pending" entries for one email_id. But I recall us having a double key where it's email_id AND some other unique-ifier. Am I misremembering? Do you even know what I'm talking about?

ctms/crud.py
```diff
@@ -267,6 +268,86 @@ def get_contacts_by_any_id(
...
+                "mofo": record.email.mofo,
+                "newsletters": record.email.newsletters,
+                "vpn_waitlist": record.email.vpn_waitlist,
+            }
+        )
+        for record in pending_records
+    ]
+
+
+def create_pending_acoustic_record(
```
jwhitlock, line 306: If I understand the design (and I'm not sure I do), there's at most one row per ``email_id``. That means if an ``email_id`` already appears in this table, the code should update the existing row. In that case, we don't need ``create_pending_acoustic_record``, since every operation is more like an upsert. ``update_pending_acoustic_record`` is closer, since it has the insert with an ``on_conflict_do_update`` clause. However, it seems overly generic for table-as-a-queue.

I think ``create_pending_acoustic_record`` and ``update_pending_acoustic_record`` need to be replaced with more specific operations:

* ``schedule_acoustic_record`` - called by the API when things change. Takes a DB session and an ``email_id``, and either creates or updates the ``PendingAcousticRecord``. I think existing rows get the retry count reset to 0 and the update timestamp set to now.
* ``retry_acoustic_record - called by the background function when Acoustic fails. Updates the retry count, leaves the rest alone.
* ``delete_acoustic_record - called by the background function when Acoustic succeeds. Similar to what you have, delete the record.

For the second two, I'd be tempted to take the ``update_timestamp`` as well, as a check that the API didn't update the row while the background job was processing it.  However, it may be better to lean on the database for row locking (such as ``SELECT FOR UPDATE`` in ``schedule_acoustic_record``).

I think this means the function interface is a DB session and and ``email_id``, and ``PendingAcousticRecordInSchema`` (and the other new schemas) are unnecessary.

I doubt I've understood the design however, since I don't have good notes from our last conversation on how this works, so I'm deferring to others.

ctms/crud.py
bsieber-mozilla: Maybe the [id](https://github.com/mozilla-it/ctms-api/pull/181/files#diff-4aa7c2a69c18e1f8b33b7cb5f54efdea916ea029661c212d3b1e4de349c91a24R195) could be used, it should be autogenerated; that or we allow a timestamp as well for this deletion. 

Also, I think this depends on how the pull is implemented...? If the pull of records from the Table contains multiple of the same email_id, wouldn't we likely pull the most up to date record when we get the [records as contacts](https://github.com/mozilla-it/ctms-api/pull/181/files#diff-45799bd240716e4cd04a9c1aeb24ad5e995dd9587db55a53d9eb85aa182d3533R288)? 

I think the edge case here is when an update happens, after we fetched the contacts, before we deleted from the table.

Which can be solved with using the same timestamp required for pulling/gets, or to provide the sequence id's (if that's possible, I can test further...)

ctms/schemas/pending_acoustic_record.py
```diff
@@ -0,0 +1,33 @@
...
+        description="Set to the retry attempt.",
+    )
+    created_date: Optional[str] = Field(
+        default=None,
+        description="ISO 8601 Timestamp with utc",
+        example="2021-01-29T18:43:49.082375+00:00",
+    )
+
+    update_timestamp: datetime = Field(
+        default_factory=datetime.utcnow,
```
jwhitlock, line 23: [datetime.utcnow](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) returns the current UTC time as a naive datetime, which is often interpreted as a local time. The Python docs recommend using [datetime.now(timezone.utc)](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) instead. This bit me about a month ago, fixed in PR #113.

I think this works (you'll need to import ``timezone`` from ``datetime`` as well):
```suggestion
        default_factory=lambda: datetime.now(timezone.utc),
```

jwhitlock (commented):
Thanks @bsieber-mozilla. I think the DB operations for this table should be tuned for queue-style operations, rather than generic CRUD operations. However, it exposes that I'm not exactly sure what happens when a record is changed twice before it is successfully synced to Acoustic. I'm hoping other have a clearer idea.

ctms/crud.py
bsieber-mozilla: `PendingAcousticRecord.update_timestamp` and 
either `PendingAcousticRecord.email_id` or `PendingAcousticRecord.id`

nice catch, thanks

ctms/schemas/pending_acoustic_record.py
bsieber-mozilla: Ah, totally makes sense, thanks! I wasn't sure how to opt for the callable, the lambda is perfect!

ctms/crud.py
bsieber-mozilla: I think your comments here are spot on. I'm hoping to make the operations more impactful for the specific tasks. 

As for the table allowing at most one row per `email_id` that is up in the air; we can make that row non-unique and allow for any number. I couldn't recall the chosen approach and opted to do it as unique for the time being, this is reversible though.

jwhitlock: I think non-unique `email_ids` would make it more like a pub-sub queue and simplify a lot. Scheduling a sync is always adding a row, and the background task is either updating or deleting rows. That reduces the need for coordination and locking bugs.

imbstack: I thought `there's at most one row per email_id` was not true. I feel like we're all on different pages for this now :slightly_frowning_face: 

I also need a bit to page this back into my head. Maybe we should chat this afternoon

ctms/crud.py
```diff
@@ -267,6 +267,63 @@ def get_contacts_by_any_id(
...
+
+def retry_acoustic_record(db: Session, pending_record: PendingAcousticRecord) -> None:
+    pending_record.retry += 1
+    pending_record.update_timestamp = datetime.now(timezone.utc)
+    # db.commit() # TODO: will this be done by the background job?
+
+
+def delete_acoustic_record(db: Session, pending_record: PendingAcousticRecord) -> None:
+    db.delete(pending_record)
+    # db.commit() # TODO: will this be done by the background job?
```
bsieber-mozilla, line 324: I'm not too sure if I need to call a commit here, or if the sessions should be committed by the background job.

Maybe just `retry` and `delete` would need this commit? And since it's only called in the background job, it could be a commit after all deletes, or after each delete?

Not sure what the preference would be here...

ctms/models.py
```diff
@@ -186,3 +187,23 @@ class MozillaFoundationContact(Base):
     mofo_relevant = Column(Boolean)
 
     email = relationship("Email", back_populates="mofo", uselist=False)
+
+
+class PendingAcousticRecord(Base):
+    __tablename__ = "pending_acoustic"
+
+    id = Column(Integer, Sequence("pending_id_sequence"), primary_key=True)
+    email_id = Column(UUID(as_uuid=True), ForeignKey(Email.email_id), nullable=False)
```
bsieber-mozilla, line 196: new shiny and no longer unique

bsieber-mozilla:
Apologies for the delay, the PR should be in a better state now and ready for a review.

Whenever you've got a chance to take a look, I've re-requested reviews. Thanks!

bsieber-mozilla:
_Also, I combined this with bstack's branch to see how the background job might look, it would appear the main thing missing is a function to push ContactSchema's to Acoustic through the api:_
draft-pr: https://github.com/mozilla-it/ctms-api/pull/187
file: https://github.com/mozilla-it/ctms-api/pull/187/files#diff-56721425dd97d7b4cd2f46cf938b7857447b6033e53803d90b5bb8bf1bc0fd6f

migrations/versions/20210505_cad3c933830b_creating_table_for_pending_acoustic_.py
```diff
@@ -0,0 +1,55 @@
...
+# revision identifiers, used by Alembic.
+revision = "cad3c933830b"  # pragma: allowlist secret
+down_revision = "bd7adf1a3801"  # pragma: allowlist secret
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Following command to create sequence added manually.
+    op.execute(CreateSequence(Sequence("pending_id_sequence")))
```
jwhitlock, line 27: Looking at the database post-migration, I think this manually created ``Sequence`` is unnecessary. Postgres includes a sequence "pending_acoustic_id_seq", and uses it for the default value of ``id`` (``nextval('pending_acoustic_id_seq'::regclass)``).

<summary>
describe commands in pgsql:
</summary>
<details>
<pre>
postgres=# \d
                   List of relations
 Schema |          Name           |   Type   |  Owner   
--------+-------------------------+----------+----------
 public | alembic_version         | table    | postgres
 public | amo                     | table    | postgres
 public | amo_id_seq              | sequence | postgres
 public | api_client              | table    | postgres
 public | emails                  | table    | postgres
 public | fxa                     | table    | postgres
 public | fxa_id_seq              | sequence | postgres
 public | mofo                    | table    | postgres
 public | mofo_id_seq             | sequence | postgres
 public | newsletters             | table    | postgres
 public | newsletters_id_seq      | sequence | postgres
 public | pending_acoustic        | table    | postgres
 public | pending_acoustic_id_seq | sequence | postgres
 public | pending_id_sequence     | sequence | postgres
 public | vpn_waitlist            | table    | postgres
 public | vpn_waitlist_id_seq     | sequence | postgres
(16 rows)

postgres=# \d pending_acoustic
                                          Table "public.pending_acoustic"
      Column      |           Type           | Collation | Nullable |                   Default                    
------------------+--------------------------+-----------+----------+----------------------------------------------
 id               | integer                  |           | not null | nextval('pending_acoustic_id_seq'::regclass)
 email_id         | uuid                     |           | not null | 
 retry            | integer                  |           | not null | 
 create_timestamp | timestamp with time zone |           | not null | now()
 update_timestamp | timestamp with time zone |           | not null | 
Indexes:
    "pending_acoustic_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "pending_acoustic_email_id_fkey" FOREIGN KEY (email_id) REFERENCES emails(email_id)
</pre>
</details>

tests/unit/test_crud.py
```diff
@@ -186,6 +192,136 @@ def test_get_contact_by_email_id_miss(dbsession):
...
+        dbsession.flush()
+
+        record_list = get_all_acoustic_records_before(
+            dbsession,
+            end_time=end_time,
+        )
+        dbsession.flush()
+
+    print(watcher.statements)
+    assert watcher.count == 4
```
jwhitlock, line 214: Nit: ``StatementWatcher`` is more useful when there is a single CRUD function being watched, or if want to check if lazy-loading is unexpectedly being used. I'm guessing there are:

* 3 SQL calls for the 3 calls to ``schedule_acoustic_record``
* 1 SQL call for ``get_all_acoustic_records_before``

The 3 insert calls are uninteresting, I wouldn't use ``StatementWatcher`` to monitor them.

I _am_ curious if accessing ``record.email`` makes a SQL call. But, that is either OK or not, depending how and when we want to get the contact data, which is a decision for the acoustic loop later.

ctms/models.py
```diff
@@ -186,3 +187,23 @@ class MozillaFoundationContact(Base):
     mofo_relevant = Column(Boolean)
 
     email = relationship("Email", back_populates="mofo", uselist=False)
+
+
+class PendingAcousticRecord(Base):
+    __tablename__ = "pending_acoustic"
+
+    id = Column(Integer, Sequence("pending_id_sequence"), primary_key=True)
```
jwhitlock, line 195: I think the ``Sequence`` can be dropped. Other models like ``Newsletter`` have an auto-created ``id`` sequence that seems to do the job. More details on the migration.

jwhitlock (approved):
Thanks @bsieber-mozilla! I think this works for populating the acoustic queue, and gets us partially to the methods needed for the sync loop. I'd like to see it merged soon. I've identified a few ignorable nits and some changes that I hope are fast.

ctms/crud.py
```diff
@@ -288,21 +288,14 @@ def get_all_acoustic_records_before(
...
-            "fxa": record.email.fxa,
-            "mofo": record.email.mofo,
-            "newsletters": record.email.newsletters,
-            "vpn_waitlist": record.email.vpn_waitlist,
-        }
-    )
-    return contact
+    contact = get_contact_by_email_id(db, record.email_id)
+    contact_schema: ContactSchema = ContactSchema.parse_obj(contact)
+    return contact_schema
```
bsieber-mozilla, line 298: This was just added as a convenience to use AcousticRecord Objects in the sync

bsieber-mozilla:
@jwhitlock, I made some changes to the tests and other areas that were mentioned in the previous review. I believe we can merge this in now if there are no remaining/tackle-able actions in this PR

bsieber-mozilla:
Merging this in

PR #185, opened 2021-05-04
Add basics of acoustic sync background job

imbstack:
This is roughly what you'll need to fit the query into. I thought about doing fancy stuff with timeouts here but I think instead we just be careful with the query and whatever is writing to acoustic to set appropriate timeouts there.

### DIFF
--- a/ctms/acoustic_sync.py
+++ b/ctms/acoustic_sync.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python3
+"""Run continuously in the background, syncing acoustic with our db."""
+
+from time import monotonic, sleep
+
+from ctms import config
+from ctms.database import get_db_engine
+
+
+def main(db, settings):
+    prev = monotonic()
+    while True:
+        print("sync here")
+        to_sleep = settings.acoustic_loop_min_secs - (monotonic() - prev)
+        if to_sleep > 0:
+            sleep(to_sleep)
+        prev = monotonic()
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


bsieber-mozilla:
I built upon your branch, and threw it here as a draft: https://github.com/mozilla-it/ctms-api/pull/187

From your branch, I rebased with my other branch that currently has the model and crud operations, so I can use the methods here as intended. 

The file you added, with some of the model additions can be seen here: https://github.com/mozilla-it/ctms-api/pull/187/files#diff-56721425dd97d7b4cd2f46cf938b7857447b6033e53803d90b5bb8bf1bc0fd6f

I believe what's left on that branch is mainly the acoustic api and how it deals with a ContactSchema object

ctms/acoustic_sync.py
```diff
@@ -0,0 +1,29 @@
...
+"""Run continuously in the background, syncing acoustic with our db."""
+
+from time import monotonic, sleep
+
+from ctms import config
+from ctms.database import get_db_engine
+
+
+def main(db, settings):
+    prev = monotonic()
```
jwhitlock, line 11: Nice, I need to remember to use ``time.monotonic()`` rather than ``time.time()`` for timing

jwhitlock (approved):
Thanks @imbstack!

I think this should live at ``ctms/bin/acoustic_sync.py`` with the other scripts, and the code that implements ``sync here`` could be a ``ctms/acoustic.py`` or similar.

PR #188, opened 2021-05-05
Adding to the background machinery

bsieber-mozilla:
Here on this PR exists an potential way to use the crud methods during the background machinery. 

- [x] The Acoustic API still has to be tackled.
- [x] And the background machinery will need to handle the `db.commit()` calls

- [ ] Secrets need to be properly provided to background job

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -1,0 +1,273 @@
+import logging
+import os
+import re
+from functools import lru_cache
+
+from django.utils.encoding import force_bytes
+from lxml import etree
+from silverpop.api import Silverpop, SilverpopResponseException
+
+from ctms import config
+from ctms.schemas import ContactSchema
+
+logger = logging.getLogger(__name__)
+
+
+@lru_cache()
+def get_settings():
+    return config.Settings()
+
+
+class AcousticResources:
+    # TODO: externalize
+    VALID_ACOUSTIC_MAIN_TABLE_FIELDS = {
+        "sfdc_id": 1,
+        "email": 1,
+        "amo_add_on_ids": 1,
+        "amo_display_name": 1,
+        "amo_email_opt_in": 1,
+        "amo_language": 1,
+        "amo_last_login": 1,
+        "amo_location": 1,
+        "amo_profile_url": 1,
+        "amo_user": 1,
+        "amo_user_id": 1,
+        "amo_username": 1,
+        "create_timestamp": 1,
+        "email_format": 1,
+        "email_id": 1,
+        "email_lang": 1,
+        "vpn_waitlist_geo": 1,
+        "vpn_waitlist_platform": 1,
+        "fxa_id": 1,
+        "fxa_account_deleted": 1,
+        "fxa_lang": 1,
+        "fxa_login_date": 1,
+        "fxa_first_service": 1,
+        "fxa_created_date": 1,
+        "fxa_primary_email": 1,
+        "double_opt_in": 1,
+        "has_opted_out_of_email": 1,
+        "mailing_country": 1,
+        "first_name": 1,
+        "last_name": 1,
+        "sumo_contributor": 1,
+        "sumo_user": 1,
+        "sumo_username": 1,
+        "basket_token": 1,
+        "unsubscribe_reason": 1,
+    }
+
+    MAIN_TABLE_SUBSCR_FLAGS = {
+        # maps the Basket/CTMS newsletter name to the Acoustic name
+        "common-voice": "sub_common_voice",
+        "mozilla-fellowship-awardee-alumni": "sub_mozilla_fellowship_awardee_alumni",
+        "firefox-accounts-journey": "sub_firefox_accounts_journey",
+        "firefox-news": "sub_firefox_news",
+        "mozilla-and-you": "sub_firefox_news",
+        "hubs": "sub_hubs",
+        "internet-health-report": "sub_internet_health_report",
+        "knowledge-is-power": "sub_knowledge_is_power",
+        "miti": "sub_miti",
+        "mixed-reality": "sub_mixed_reality",
+        "mozilla-festival": "sub_mozilla_festival",
+        "mozilla-foundation": "sub_mozilla_foundation",
+        "mozilla-technology": "sub_mozilla_technology",
+        "mozillians-nda": "sub_mozillians_nda",
+        "take-action-for-the-internet": "sub_take_action_for_the_internet",
+        "test-pilot": "sub_test_pilot",
+        "about-mozilla": "sub_about_mozilla",
+        "app-dev": "sub_apps_and_hacks",
+    }
+
+
+class Acoustic(Silverpop):
+    """
+    :param client_id
+    :param client_secret
+    :param refresh_token
+    :param server_number
+    """
+
+    @staticmethod
+    def _process_response(resp):
+        logger.debug("Response: %s", resp.text)
+        response = etree.fromstring(resp.text.encode("utf-8"))
+        # success = response.find(".//SUCCESS")
+
+        # how main table failures are reported:
+        fault = response.find(".//Fault/FaultString")
+        if fault is not None:
+            print(fault.text)
+            raise SilverpopResponseException(fault.text)
+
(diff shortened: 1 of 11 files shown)


ctms/acoustic_sync.py
```diff
@@ -1,10 +1,48 @@
...
+    contact: ContactSchema = get_acoustic_record_as_contact(db, pending_record)
+    is_success = sync_contact_with_acoustic(contact)
+
+    if is_success:
+        delete_acoustic_record(db, pending_record)
+    else:
+        retry_acoustic_record(db, pending_record)
+
+
+def sync_records(db):
```
bsieber-mozilla, line 40: calling this method at this print:
https://github.com/mozilla-it/ctms-api/pull/188/files#diff-56721425dd97d7b4cd2f46cf938b7857447b6033e53803d90b5bb8bf1bc0fd6fR51

ctms/acoustic_service.py
```diff
@@ -0,0 +1,273 @@
...
+        response = self.session.post(self.api_endpoint, data={"xml": force_bytes(xml)})
+        return self._process_response(response)
+
+
+class CTMSToAcousticService:
+    def __init__(
+        self,
+        client_id=os.environ["ACOUSTIC_CLIENT_ID"],
+        client_secret=os.environ["ACOUSTIC_CLIENT_SECRET"],
+        refresh_token=os.environ["ACOUSTIC_REFRESH_TOKEN"],
```
bsieber-mozilla, line 125: expected secrets embedded in env vars, currently. 

What is the approach for secrets on background machinery?

ctms/acoustic_service.py
```diff
@@ -0,0 +1,273 @@
...
+
+from ctms import config
+from ctms.schemas import ContactSchema
+
+logger = logging.getLogger(__name__)
+
+
+@lru_cache()
+def get_settings():
+    return config.Settings()
```
bsieber-mozilla, line 18: used below for config env vars:

https://github.com/mozilla-it/ctms-api/pull/188/files#diff-4b3cc193cf9732b1c5ea7f0e45a90c6c9234db1217ae97bc831babc0c505eb7dR267

https://github.com/mozilla-it/ctms-api/pull/188/files#diff-4b3cc193cf9732b1c5ea7f0e45a90c6c9234db1217ae97bc831babc0c505eb7dR260

ctms/bin/acoustic_sync.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+"""Run continuously in the background, syncing acoustic with our db."""
```
bsieber-mozilla, line 2: moved to `ctms/bin`

ctms/config.py
```diff
@@ -13,6 +13,8 @@ class Settings(BaseSettings):
     logging_level: Literal["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"] = "INFO"
     sentry_debug: bool = False
     acoustic_loop_min_secs: int = 5
+    acoustic_main_table_id: int = 1390189
+    acoustic_newsletter_table_id: int = 1390247
```
bsieber-mozilla, line 17: Added env var configuration values here.

ctms/acoustic_service.py
```diff
@@ -0,0 +1,273 @@
+import logging
```
bsieber-mozilla, line 1: c+v from https://github.com/cvalaas/acoustic-api-test/blob/main/test_acoustic_call.py

bsieber-mozilla: _c+v being `chris+valaas` or `ctrl-c+ctrl-v`_

bsieber-mozilla:
### TODO:

- [ ] Testing
- [ ] Determine strategy for secrets

ctms/acoustic_service.py
cvalaas: I was wondering.

bsieber-mozilla:
# New Env Vars Added
## Secrets
client_id=os.environ["ACOUSTIC_CLIENT_ID"],
client_secret=os.environ["ACOUSTIC_CLIENT_SECRET"],
refresh_token=os.environ["ACOUSTIC_REFRESH_TOKEN"],
## Config (in Settings)
CTMS_acoustic_main_table_id
CTMS_acoustic_newsletter_table_id

ctms/acoustic_service.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    :param client_id
+    :param client_secret
+    :param refresh_token
+    :param server_number
+    """
+
+    @staticmethod
+    def _process_response(resp):
+        logger.debug("Response: %s", resp.text)
```
imbstack, line 94: I'm thinking we should avoid logging responses/requests since they'll include PII in all likelihood

ctms/acoustic_service.py
```diff
@@ -0,0 +1,273 @@
...
+                    "update_date": self._transform_field_for_acoustic(
+                        newsletter["update_timestamp"]
+                    ),
+                }
+            )
+            # and finally flip the main table's sub_<newsletter> flag to true for each subscription
+            if newsletter["subscribed"]:
+                acoustic_main_table[
+                    AcousticResources.MAIN_TABLE_SUBSCR_FLAGS[newsletter["name"]]
+                ] = "1"
```
imbstack, line 208: this won't necessarily always be true in ctms I think, right? This was true in an initial sync out of bq but not in the general case

ctms/acoustic_service.py
```diff
@@ -0,0 +1,273 @@
...
+            return "0"
+        # Acoustic doesn't have timestamps, so make timestamps into dates.
+        # Is this appropriate for all TS fields?
+        # Is the below RE appropriate for what we'll actually be getting from Postgres?
+        try:
+            matching = re.search(r"^(\d{4}-\d{2}-\d{2})T\d", data)
+            if matching:
+                return matching.groups()[0]
+        except TypeError:
+            pass
```
imbstack, line 226: I'm a bit confused by this. We're operating on data we got out of postgres, right? So this should be able to be operated on as whatever sqlalchemy gives us for a datetime? We can transform it to whatever string acoustic wants without the use of regex (unless I am not understanding what's going on here)

imbstack (commented):
Just some initial thoughts! Looking good so far

ctms/acoustic_service.py
cvalaas: I don't understand this comment. _What_ won't necessarily be true in ctms?

ctms/acoustic_service.py
cvalaas: sounds true, yeah.

ctms/acoustic_service.py
imbstack: "flip the main table's sub_<newsletter> flag to true for each subscription"

won't we have "subscribed: False" in our tables?

cvalaas: Yes, and those won't get flipped. At least, assuming that field as given to us as a boolean. If it's just a string, then that'd be trouble.

ctms/acoustic_service.py
bsieber-mozilla: Attempted to address this in new commits

ctms/sync.py
```diff
@@ -0,0 +1,66 @@
...
+    def sync_records(self, db, end_time=datetime.now(timezone.utc)):
+        # Get all Records before current time
+        all_acoustic_records_before_now: List[
+            PendingAcousticRecord
+        ] = get_all_acoustic_records_before(db, end_time=end_time)
+
+        # For each record, attempt downstream sync; commit changes to db
+        for acoustic_record in all_acoustic_records_before_now:
+            self._sync_pending_record(db, acoustic_record)
+            db.commit()
```
bsieber-mozilla, line 66: `db.commit()` handled here, so background machinery just calls this classes method.

in ref to TODO:
https://github.com/mozilla-it/ctms-api/pull/188/commits/ee8a37d9922cc136a41e20af3aa128f72f0383b3#diff-45799bd240716e4cd04a9c1aeb24ad5e995dd9587db55a53d9eb85aa182d3533R319

PR #193, opened 2021-05-07
Acoustic Sync - From Ctms/DB to Acoustic

bsieber-mozilla:
Starting from @cvalaas 's acoustic code, did some refactoring and testing.

Draft Ref #188


ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+        "take-action-for-the-internet": "sub_take_action_for_the_internet",
+        "test-pilot": "sub_test_pilot",
+        "about-mozilla": "sub_about_mozilla",
+        "app-dev": "sub_apps_and_hacks",
+        "firefox-welcome": "firefox-welcome",  # TODO: Confirm approach, found newsletters through testing
+        "mozilla-welcome": "mozilla-welcome",  # TODO: Confirm
+        "firefox-os": "firefox-os",  # TODO: Confirm
+        "ambassadors": "ambassadors",  # TODO: Confirm
+        "maker-party": "maker-party",  # TODO: Confirm
+        "mozilla-learning-network": "mozilla-learning-network",  # TODO: Confirm
```
cvalaas, line 79: If these were newsletters that were still active, then the Acoustic name would be "sub__something_".
If we need these in here for testing, then I'd change the Acoustic name to something like "FAKE_FOR_TESTING" so it's clear we're not actually using them. Or something?
I dunno, regardless, they shouldn't be passed on to Acoustic, since Acoustic doesn't know about them.

ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+                    # print(f"Skipping CTMS field ({contact_attr}, {inner_attr}) because no match "
+                    #       f"in Acoustic") # TODO: Convert to logging
+        return acoustic_main_table
+
+    def _newsletter_converter(self, acoustic_main_table, contact):
+        # create the RT rows for the newsletter table in acoustic
+        newsletter_rows = []
+        contact_newsletters: List[NewsletterSchema] = contact.newsletters
+        for newsletter in contact_newsletters:
+            if newsletter.name in AcousticResources.MAIN_TABLE_SUBSCR_FLAGS.keys():
```
cvalaas, line 194: I wonder if we should log unhandled newsletters (ie- the "else" case here)?

cvalaas: not urgent

ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+                    "newsletter_lang": newsletter.lang,
+                    "newsletter_source": newsletter.source,
+                    "newsletter_unsub_reason": newsletter.unsub_reason,
+                    "create_date": create_timestamp,
+                    "update_date": update_timestamp,
+                }
+                newsletter_rows.append(newsletter_template)
+                # and finally flip the main table's sub_<newsletter> flag to true for each subscription
+                if newsletter.subscribed:
+                    acoustic_main_table[
```
cvalaas, line 215: is "acoustic_main_table" an accessible dict in this def?

cvalaas: I guess it is, huh. But it doesn't get passed back out. Is that weird?

cvalaas (changes requested):
The first comment (about the fields in MAIN_TABLE_SUBSCR_FLAGS) is the only one that needs action, I think

ctms/acoustic_service.py
bsieber-mozilla: Ok! I will do some testing modification. Just wanted to make sure this data-set of flags was up to date, was hoping to confirm/deny!

ctms/acoustic_service.py
bsieber-mozilla: It was provided as a mutable dictionary... but yeah, that could be a bit weird. I can modify the behavior to return both dictionaries

ctms/acoustic_service.py
bsieber-mozilla: I wasn't sure how logging is handled with the background jobs... I can throw some basic logging in there on next update-commit.

ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+                        AcousticResources.MAIN_TABLE_SUBSCR_FLAGS[newsletter.name]
+                    ] = "1"
+            # else:
+            # print(f"Skipping Newsletter {newsletter.name} because no match in Acoustic") # TODO: Convert to logging
```
bsieber-mozilla, line 219: _(Commented out in ref to adding logging)_

tests/unit/test_acoustic_sync.py
```diff
@@ -0,0 +1,128 @@
...
+CTMS_ACOUSTIC_MAIN_TABLE_ID = "1"
+CTMS_ACOUSTIC_NEWSLETTER_TABLE_ID = "9"
+
+
+@pytest.fixture()
+def no_acoustic():
+    patcher = mock.patch("ctms.acoustic_service.Acoustic")
+    patcher.start()
+    yield patcher
+    patcher.stop()
```
jwhitlock, line 21: I think the purpose is to mock ``Acoustic`` so it doesn't call anything, but it looks like the mock is unused in the code, so it doesn't really need to be returned. It could be moved to ``base_ctms_acoustic_service``,  something like:

```python
def base_ctms_acoustic_service():
    with mock.patch("ctms.acoustic_service.Acoustic"):
        yield acoustic_service.CTMSToAcousticService(
            client_id=CTMS_ACOUSTIC_CLIENT_ID,
            client_secret=CTMS_ACOUSTIC_CLIENT_SECRET,
            refresh_token=CTMS_ACOUSTIC_REFRESH_TOKEN,
            acoustic_main_table_id=CTMS_ACOUSTIC_MAIN_TABLE_ID,
            acoustic_newsletter_table_id=CTMS_ACOUSTIC_NEWSLETTER_TABLE_ID,
            server_number=6,
        )
```

ctms/sync.py
```diff
@@ -0,0 +1,66 @@
...
+    def sync_records(self, db, end_time=datetime.now(timezone.utc)):
+        # Get all Records before current time
+        all_acoustic_records_before_now: List[
+            PendingAcousticRecord
+        ] = get_all_acoustic_records_before(db, end_time=end_time)
+
+        # For each record, attempt downstream sync; commit changes to db
+        for acoustic_record in all_acoustic_records_before_now:
+            self._sync_pending_record(db, acoustic_record)
+            db.commit()
```
imbstack, line 66: Do we actually want to commit for every record or just after they're all processed? If this works now, we can land it as-is but worth looking into how we want to do these transactions

ctms/bin/acoustic_sync.py
```diff
@@ -1,16 +1,24 @@
...
-
 from time import monotonic, sleep
 
 from ctms import config
 from ctms.database import get_db_engine
+from ctms.sync import CTMSToAcousticSync
 
 
 def main(db, settings):
+    sync_service = CTMSToAcousticSync(
```
imbstack, line 11: So this is where my acoustic knowledge breaks down:

We get a refresh token from them that doesn't change. Does the pysilverpop client handle refreshing our creds in the background or will this one we construct outside of the loop eventually go out of date?

imbstack (approved):
some questions but worth landing as-is to unblock testing I expect

tests/unit/test_sync.py
```diff
@@ -0,0 +1,97 @@
...
+from unittest import mock
+from unittest.mock import MagicMock
+
+import pytest
+
+from ctms.crud import schedule_acoustic_record
+from ctms.sync import CTMSToAcousticSync
+from tests.unit.test_crud import StatementWatcher
+
+CTMS_ACOUSTIC_CLIENT_ID = "CLIENT"
```
jwhitlock, line 11: These test-only constants are just used in ``CTMSToAcousticSync`` below, I think they can be used directly there, unless there is more planned in the future that may use them

ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+        client_secret,
+        refresh_token,
+        acoustic_main_table_id,
+        acoustic_newsletter_table_id,
+        server_number=6,
+    ):
+        """
+        Construct a CTMSToAcousticService object that can interface between contacts and acoustic forms of data
+        """
+        self.acoustic = Acoustic(
```
jwhitlock, line 130: Nit:I think a cleaner interface would be to pass an ``Acoustic`` instance to ``CTMSToAcousticService``, rather than to construct it in the ``__init__`` method. For example, it would allow you to pass a mock object for tests.

ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+            client_id=client_id,
+            client_secret=client_secret,
+            refresh_token=refresh_token,
+            server_number=server_number,
+        )
+        if isinstance(acoustic_main_table_id, str):
+            acoustic_main_table_id = int(acoustic_main_table_id)
+        self.acoustic_main_table_id = acoustic_main_table_id
```
jwhitlock, line 138: If an int is required, this should do it in one line:
```suggestion
        self.acoustic_main_table_id = int(acoustic_main_table_id)
```

ctms/config.py
```diff
@@ -12,7 +12,15 @@ class Settings(BaseSettings):
...
         env_prefix = "ctms_"
+
+
+class BackgroundSettings(Settings):
+    acoustic_loop_min_secs: int = 5
+    acoustic_client_id: str
+    acoustic_client_secret: str
+    acoustic_refresh_token: str
+    acoustic_main_table_id: str
+    acoustic_newsletter_table_id: str
```
jwhitlock, line 26: If these are supposed to be ints, then they should be:
```suggestion
    acoustic_main_table_id: int
    acoustic_newsletter_table_id: int
```

jwhitlock (commented):
Thanks @bsieber-mozilla! I had time for a quick code review, but I'm deferring to others with Acoustic experience for approval. Most of my comments are nits, and can be ignored in the interest in getting the first version deployed so we have a chance to fix the first round of problems.

ctms/bin/acoustic_sync.py
bsieber-mozilla: From my understanding (imparted by @cvalaas, so correct me if I'm wrong!)
... is that the token is not expected to change and should not need to be updated.

ctms/acoustic_service.py
```diff
@@ -0,0 +1,279 @@
...
+        if failures is not None:
+            for fail in failures:
+                raise SilverpopResponseException(etree.tostring(fail))
+
+        return response
+
+    def _call(self, xml):
+        # Following basket's suggested pattern,
+        # adding force_bytes to deal with some Acoustic rejections
+        response = self.session.post(self.api_endpoint, data={"xml": force_bytes(xml)})
```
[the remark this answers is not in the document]
bsieber-mozilla: Thanks for the cherry-pick suggestion; I was tying to figure out if it was possible to just grab the encoding library but to no luck.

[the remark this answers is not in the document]
bsieber-mozilla: As for FastAPI to Django 🤣   
I'd be up for it, but might not be best to do at this time  🤷

ctms/bin/acoustic_sync.py
jwhitlock: They use [OAuth2Session](https://github.com/theatlantic/pysilverpop/blob/ee9d60a6e99d36f123729746ceba6003ac56cb59/silverpop/api.py#L179-L183), in a similar way to Basket when it calls CTMS. I've read that code, the class takes care of storing any changed refresh token.

tests/unit/test_acoustic_sync.py
```diff
@@ -0,0 +1,128 @@
+import datetime
```
bsieber-mozilla, line 1: dupe --remove

bsieber-mozilla: I might've left in this in a tired state, when I was renaming things...

ctms/bin/acoustic_sync.py
jwhitlock: and, like CTMS, the refresh URL is also the token URL, and you pass in the client ID and secret, so I'm guessing Acoustic's token granting implementation is very close to CTMS's, and the refresh token is nonsense.

ctms/acoustic_service.py
bsieber-mozilla: I believe in the recent commit: c42ccef86b1835e4a6b337946166ca09827ccde7

I made the changes requested; however, I feel that I need to still tweak the behavior of the way I'm using mocks in tests.

ctms/acoustic_service.py
bsieber-mozilla: Added some basic logging, please re-review time-permitting. (resolving this, opening a new one would be easiest for me)

bsieber-mozilla:
(@cvalaas, @imbstack, @jwhitlock)
I believe I've addressed the majority of the comments above; if there are any other requested changes or nit's, please call them out. 

_I'm ready to merge this in if no other comments_

imbstack (approved):
Let's land it!

ctms/bin/acoustic_sync.py
cvalaas: I know the refresh token is required, but if it changes (I expect it does during a session), then I would expect pysilverpop to handle that.

cvalaas (approved):

jwhitlock (approved):

bsieber-mozilla:
🚢 !

PR #202, opened 2021-05-11
Adding ability to use env var for retry limit

bsieber-mozilla:


ctms/config.py
```diff
@@ -28,6 +28,8 @@ class Config:
 
 
 class BackgroundSettings(Settings):
+    acoustic_retry_limit: int = 6
+    acoustic_server_number: int = 6
```
bsieber-mozilla, line 32: CTMS_ACOUSTIC_RETRY_LIMIT
CTMS_ACOUSTIC_SERVER_NUMBER

Two potential new env vars for tuning the background machinery.

jwhitlock (approved):

Issue #165, opened 2021-04-28
Allow HEAD to ``/__heartbeat__``, ``/__lb_heartbeat__`` endpoints


jwhitlock:
Allow ``HEAD`` as well as ``GET`` requests to [Dockerflow](https://github.com/mozilla-services/Dockerflow) endpoints:

* ``/__heartbeat__``
* ``/__lb_heartbeat__``

A user or service in production is requesting ``HEAD /__heartbeat__``, and getting a 405. It is then trying ``GET /__heartbeart__`` ``HEAD`` makes sense, since the status code of ``200`` is sufficient to determine that the service is healthy.

bsieber-mozilla:
I'm closing this since #205 has been merged in.

Issue #153, opened 2021-04-26
Mark log entries as ignorable


jwhitlock:
When viewing logs, it is useful to ignore some entries:

* Log entries generated by Kubernetes liveness checks
* Redirects from the root URL to `/docs`

If these were marked in the logs, such as with ``"trivial": true``, then we could filter them with a simple Papertrail text filter (``-"\"trivial\": true"``), rather than exclude each endpoint.

jwhitlock:
Other possible names, "minor", "noise", "jerkwater"

jwhitlock:
Taking this now. "trivial" sounds good to me after a few weeks

The ones I'm marking in this first pass:

* Any request to the homepage ``/`` that results in a 307 redirect to ``/docs``. Any client details will be on that second request.
* When the new relic client calls ``/__heartbeat__`` and gets a 200
* When the kubernetes probe call ``/__lbheartbeat__`` and gets a 200
* When the prometheus client calls ``/metrics`` and gets a 200

jwhitlock:
Deployed, working in stage.

PR #205, opened 2021-05-11
Adding HEAD for heartbeat api calls. Issue #165

bsieber-mozilla:
Allow HEAD as well as GET requests to :

    /__heartbeat__
    /__lb_heartbeat__

### DIFF
--- a/ctms/sample_data.py
+++ b/ctms/sample_data.py
@@ -177,7 +177,8 @@ class ContactVendor:
     }
 
     def __getitem__(self, key: str) -> ContactSchema:
-        return self.contacts[key][0].copy(deep=True)
+        contact: ContactSchema = self.contacts[key][0].copy(deep=True)
+        return contact
 
     def get_not_written(self, key: str) -> Set[str]:
         return self.contacts[key][1]
--- a/tests/unit/test_api.py
+++ b/tests/unit/test_api.py
@@ -402,6 +402,10 @@ API_TEST_CASES: Tuple[Tuple[str, str, Any], ...] = (
         "/updates?start=2020-01-22T03%3A24%3A00%2B00%3A00&end=&limit=&after=&mofo_relevant=",
         None,
     ),
+    ("HEAD", "/__lbheartbeat__", None),
+    ("GET", "/__lbheartbeat__", None),
+    ("HEAD", "/__heartbeat__", None),
+    ("GET", "/__heartbeat__", None),
 )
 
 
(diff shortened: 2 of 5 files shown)


ctms/sample_data.py
```diff
@@ -177,7 +177,8 @@ class ContactVendor:
     }
 
     def __getitem__(self, key: str) -> ContactSchema:
-        return self.contacts[key][0].copy(deep=True)
+        contact: ContactSchema = self.contacts[key][0].copy(deep=True)
+        return contact
```
bsieber-mozilla, line 181: > ctms/sample_data.py:180: error: Returning Any from function declared to return "ContactSchema"
> Found 1 error in 1 file (checked 2 source files)
Fix for this mypy pre-commit issue which came up?

tests/unit/test_metrics.py
```diff
@@ -15,10 +15,10 @@
 # Higher numbers = more ways to slice data, more storage, more processing time for summaries
 
 # Cardinality of ctms_requests_total counter
-METHOD_PATH_CODE_COMBINATIONS = 45
+METHOD_PATH_CODE_COMBINATIONS = 47
 
 # Cardinality of ctms_requests_duration_seconds histogram
-METHOD_PATH_CODEFAM_COMBOS = 31
+METHOD_PATH_CODEFAM_COMBOS = 33
 DURATION_BUCKETS = 8
```
bsieber-mozilla, line 22: With 2 new HEAD methods, is this correctly adjusted...?

tests/unit/test_api.py
```diff
@@ -410,30 +414,47 @@ def test_unauthorized_api_call_fails(
...
+        "GET": ({"method": "GET", "url": path, "params": params}, {200}),
+        "HEAD": ({"method": "HEAD", "url": path}, {200}),
+        "PUT": (put_patch_post("PUT"), {200, 201}),  # Either creates or updates
+        "PATCH": (put_patch_post("PATCH"), {200}),
+        "POST": (put_patch_post("POST"), {201}),
+    }
+    if method in _map.keys():
+        _request, _results = _map.get(method)
+        resp = client.request(**_request)
+        assert resp.status_code in _results
```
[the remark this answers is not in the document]
jwhitlock: Yes, we can split up this file. I hit this warning when adding ``PATCH``, so I put all those tests in ``test_api_patch.py``. A similar method could be used to break out the ``POST`` and ``PUT`` tests into their own files.

I've been testing monitoring endpoints in ``test_app.py``, I think these new checks should go there as well, which will keep us under 1000 for another day.

tests/unit/test_metrics.py
jwhitlock: That appears correct to me, adds two new method/path combos

ctms/sample_data.py
jwhitlock: Maybe! I don't see this error in my dev environment, however, so I can't tell for sure.

The sample data needs to be moved out of ``ctms`` and into ``tests/unit``, but that's for a future PR.

jwhitlock (approved):
Thanks @bsieber-mozilla! I think the code change looks good, but the tests should be in ``tests_app.py`` or a new file ``tests_monitor.py``, not in ``tests_api.py``. 

There's some stalled work to add this to FastAPI in https://github.com/tiangolo/fastapi/issues/1773. I don't think we should wait for it.

tests/unit/test_monitor.py
```diff
@@ -0,0 +1,38 @@
...
+def test_authorized_api_call_succeeds(client, example_contact, method, path, params):
+    """Calling the API without credentials fails."""
+    _map = {  # METHOD: ( {_REQUEST_PARAMS} , {_EXPECTED_RESULTS}
+        "GET": ({"method": "GET", "url": path, "params": params}, {200}),
+        "HEAD": ({"method": "HEAD", "url": path}, {200}),
+    }
+    if method in _map.keys():
+        _request, _results = _map.get(method)
+        resp = client.request(**_request)
+        assert resp.status_code in _results
```
jwhitlock, line 38: No, this copies the API tests without making any of the changes needed to test the monitoring endpoints. The test is called ``_fails`` but tests that they return 200s, tests with an authenticated client even though API credentials are not needed, and for some reason passes ``{"params": None}`` for a few cases.

Let's do this:

```suggestion
@pytest.mark.parametrize("method", ("HEAD", "GET"))
@pytest.mark.parametrize("path", ("/__lbheartbeat", "/__heartbeat"))
def test_monitoring_endpoints(anon_client, method, path):
    """Monitoring endpoints can be called without credentials"""
    resp = anon_client.request(method, path)
    assert resp.status_code == 200
```

jwhitlock: but, the existing tests do pass and test the new methods.

I'm touching the monitoring tests for issue #153, I can make my changes then if you want to merge and move on to the next issue.

tests/unit/test_monitor.py
```diff
@@ -2,37 +2,9 @@
 
 import pytest
```
bsieber-mozilla, line 3: _Going to hop to branch to clean this_

bsieber-mozilla: from `db_session` to `dbsession`

bsieber-mozilla:
@jwhitlock this should resolve, then I'll merge it in

PR #207, opened 2021-05-11
Adding FeatureFlags to the Sync Job as Background EnvVars

bsieber-mozilla:
```
CTMS_ACOUSTIC_SYNC_FEATURE_FLAG
CTMS_ACOUSTIC_INTEGRATION_FEATURE_FLAG
```

Both booleans, defaulted to False in the BackgroundSettings.

`CTMS_ACOUSTIC_SYNC_FEATURE_FLAG` allows for enabling (True) the background machinery to begin looping and pulling records, otherwise (False) the background machinery will not loop.  

`CTMS_ACOUSTIC_INTEGRATION_FEATURE_FLAG` allows for enabling (True) the records to be sent to acoustic, otherwise (False) records will be dropped from the database.

---
```
CTMS_ACOUSTIC_SYNC_FEATURE_FLAG=True
CTMS_ACOUSTIC_INTEGRATION_FEATURE_FLAG=True 
```

Is the expected end state, when we're pulling all records and syncing them into acoustic.

---
```
CTMS_ACOUSTIC_SYNC_FEATURE_FLAG=True
CTMS_ACOUSTIC_INTEGRATION_FEATURE_FLAG=False 
```

Is an interim state, when we're pulling all records without pushing to acoustic, marking them as successful to delete the pending records.

 ---
```
CTMS_ACOUSTIC_SYNC_FEATURE_FLAG=False
CTMS_ACOUSTIC_INTEGRATION_FEATURE_FLAG=True/False 
```

A shutoff-state, when `CTMS_ACOUSTIC_SYNC_FEATURE_FLAG` is False, the system will not run, no matter the `CTMS_ACOUSTIC_INTEGRATION_FEATURE_FLAG`

### DIFF
--- a/ctms/sync.py
+++ b/ctms/sync.py
@@ -56,16 +56,16 @@ class CTMSToAcousticSync:
     def _sync_pending_record(self, db, pending_record: PendingAcousticRecord):
         try:
             if self.is_acoustic_enabled:
+                contact: ContactSchema = get_acoustic_record_as_contact(
+                    db, pending_record
+                )
+                is_success = self.sync_contact_with_acoustic(contact)
+            else:
                 self.logger.debug(
                     "Acoustic is not currently enabled. Records will be classified as successful and "
                     "dropped from queue at this time."
                 )
                 is_success = True
-            else:
-                contact: ContactSchema = get_acoustic_record_as_contact(
-                    db, pending_record
-                )
-                is_success = self.sync_contact_with_acoustic(contact)
 
             if is_success:
                 # on success delete pending_record from table
--- a/ctms/sync.py
+++ b/ctms/sync.py
@@ -48,21 +48,24 @@ class CTMSToAcousticSync:
         """
         try:
             # Convert ContactSchema to Acoustic Readable, attempt API call
-            if self.is_acoustic_enabled:
-                return self.ctms_to_acoustic.attempt_to_upload_ctms_contact(contact)
-            self.logger.debug(
-                "Acoustic is not currently enabled. Records will be classified as successful and "
-                "dropped from queue at this time."
-            )
-            return True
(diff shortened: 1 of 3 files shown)


bsieber-mozilla:
_Other approaches are encouraged!_

imbstack (approved):
really nice approach!

ctms/sync.py
```diff
@@ -48,21 +48,24 @@ def sync_contact_with_acoustic(self, contact: ContactSchema):
...
+                self.logger.debug(
+                    "Acoustic is not currently enabled. Records will be classified as successful and "
+                    "dropped from queue at this time."
+                )
+                is_success = True
+            else:
+                contact: ContactSchema = get_acoustic_record_as_contact(
+                    db, pending_record
+                )
+                is_success = self.sync_contact_with_acoustic(contact)
```
bsieber-mozilla, line 68: Moving the FF check to here. Removing the need to call the db for contact data and saving db calls.

ctms/sync.py
```diff
@@ -53,8 +55,17 @@ def sync_contact_with_acoustic(self, contact: ContactSchema):
...
+            else:
+                self.logger.debug(
+                    "Acoustic is not currently enabled. Records will be classified as successful and "
+                    "dropped from queue at this time."
+                )
+                is_success = True
```
bsieber-mozilla, line 69: _acoustic FF check here_

If enabled continue w/ integration path; else, skip pulling contact and syncing, send success/True signal to delete records.

bsieber-mozilla:
When we merge this, we may want to fast-follow a PR for adding in the env vars to the deploy. 

_(Re-enabling stage, disabling prod: https://github.com/mozilla-it/ctms-infra/pull/36)_

duallain (approved):
LGTM, thanks for adding a feature flag.

Issue #182, opened 2021-05-04
Handle invalid `after` parameter in `/updates`


jwhitlock:
A caller to `/updates` passed `after=null`, which is probably a mistake. This is valid base64, and decodes as `'\x9e\xe9e'`, but does not pass UTF-8 decoding, and raises an exception. This, and other invalid `after` values, should instead return an error with a 4xx status code.

bsieber-mozilla:
a query param with the value of `null` should no longer cause any errors. Although, when a "string" of "null" is seen, a decode will be attempted, and should fail. This failure indicates that the "after" string is likely not a value the compressed in a way the system understands, and is a client err for a misformatted 'after' param.

PR #208, opened 2021-05-11
Updating to 4xx on bad requests to bulk read

bsieber-mozilla:
On query params, a new validation error of 422 will be presented to users hitting the endpoint with any string.:
- empty string "" as an exception
- after is allowed to have strs, decoding is wrapped to throw a 422 if after is invalid.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -1,12 +1,10 @@
-import base64
 import sys
 import time
 from datetime import datetime, timedelta
 from functools import lru_cache
-from typing import Dict, List, Literal, Optional, Tuple, Union
+from typing import Dict, List, Literal, Optional, Union
 from uuid import UUID, uuid4
 
-import dateutil.parser
 import sentry_sdk
 import structlog
 import uvicorn
@@ -240,23 +238,6 @@ def get_contacts_by_ids(
     ]
 
 
-def extractor_for_bulk_encoded_details(after: str) -> Tuple[str, datetime]:
-    str_decode = base64.urlsafe_b64decode(after)
-    result_after_list = str(str_decode.decode("utf-8")).split(",")
-    after_email_id = result_after_list[0]
-    after_start_time = dateutil.parser.parse(result_after_list[1])
-    return after_email_id, after_start_time
-
-
-def compressor_for_bulk_encoded_details(last_result: CTMSResponse):
-    last_email_id = last_result.email.email_id
-    last_update_time = last_result.email.update_timestamp
-    result_after_encoded = base64.urlsafe_b64encode(
-        f"{last_email_id},{last_update_time}".encode("utf-8")
-    )
-    return result_after_encoded.decode()
-
-
 def get_bulk_contacts_by_timestamp_or_4xx(
     db: Session,
     start_time: datetime,
@@ -269,9 +250,10 @@ def get_bulk_contacts_by_timestamp_or_4xx(
     after_email_id = None
     after_start_time = start_time
     if after is not None:
-        after_email_id, after_start_time = extractor_for_bulk_encoded_details(
-            after=after
-        )
+        (
+            after_email_id,
+            after_start_time,
+        ) = BulkRequestSchema.extractor_for_bulk_encoded_details(after)
 
     results = get_bulk_contacts(
         db=db,
@@ -302,7 +284,11 @@ def get_bulk_contacts_by_timestamp_or_4xx(
         next_url = None
     else:
         last_result: CTMSResponse = results[-1]
-        after_encoded = compressor_for_bulk_encoded_details(last_result)
+        after_encoded = BulkRequestSchema.compressor_for_bulk_encoded_details(
+            last_email_id=last_result.email.email_id,
+            last_update_time=last_result.email.update_timestamp,
+        )
+
         next_url = (
             f"{get_settings().server_prefix}/updates?"
             f"start={start_time.isoformat()}"
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -1,7 +1,7 @@
 import base64
 import sys
 import time
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 from functools import lru_cache
 from typing import Dict, List, Literal, Optional, Tuple, Union
 from uuid import UUID, uuid4
@@ -53,6 +53,7 @@ from .schemas import (
     AddOnsSchema,
     ApiClientSchema,
     BadRequestResponse,
+    BulkRequestSchema,
     ContactInSchema,
     ContactPatchSchema,
     ContactPutSchema,
@@ -239,17 +240,11 @@ def get_contacts_by_ids(
 
 
 def extractor_for_bulk_encoded_details(after: str) -> Tuple[str, datetime]:
-    try:
-        str_decode = base64.urlsafe_b64decode(after)
-        result_after_list = str(str_decode.decode("utf-8")).split(",")
-        after_email_id = result_after_list[0]
-        after_start_time = dateutil.parser.parse(result_after_list[1])
-        return after_email_id, after_start_time
-    except Exception as e:
-        raise HTTPException(
-            status_code=422,
-            detail="Error in Client Request: 'after' param validation error.",
-        ) from e
+    str_decode = base64.urlsafe_b64decode(after)
+    result_after_list = str(str_decode.decode("utf-8")).split(",")
+    after_email_id = result_after_list[0]
+    after_start_time = dateutil.parser.parse(result_after_list[1])
+    return after_email_id, after_start_time
 
 
 def compressor_for_bulk_encoded_details(last_result: CTMSResponse):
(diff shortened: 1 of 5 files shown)


ctms/app.py
```diff
@@ -255,7 +261,7 @@ def compressor_for_bulk_encoded_details(last_result: CTMSResponse):
     return result_after_encoded.decode()
 
 
-def get_bulk_contacts_by_timestamp(
+def get_bulk_contacts_by_timestamp_or_404(
```
bsieber-mozilla, line 264: _or_4XX_

bsieber-mozilla, line 264:
```suggestion
def get_bulk_contacts_by_timestamp_or_4XX(
```

ctms/app.py
```diff
@@ -320,7 +326,7 @@ def get_bulk_contacts_by_timestamp(
 
 
 def updates_helper(value, default):
-    blank_vals = ["", None]
+    blank_vals = [None, "", "null"]
```
bsieber-mozilla, line 329: adding "null" as a default param, that should be converted to an int, 5.

bsieber-mozilla:
Issues in reference:
#140 
#182

bsieber-mozilla:
```
    start: datetime,
    end: Optional[Union[datetime, Literal[""]]] = None,
    limit: Optional[Union[int, Literal[""]]] = None,
    after: Optional[str] = None,
    mofo_relevant: Optional[Union[bool, Literal[""]]] = None,
```

`start` is the only expected query parameter.

`end` can be provided as a datetime or "" (if "", default to _now_) or not provided

`limit` can be sent as an int or "" (if "", default to 10) or not provided

`after` can be sent as a base64 str (in valid decode-able format) or "" (if "", default to None) or not provided

`mofo_relevant` can be sent as a boolean or "" (if "", default to None) or not provided

ctms/app.py
```diff
@@ -625,25 +619,21 @@ def partial_update_ctms_contact(
...
-        end_time=end_param,
-        after=after_param,
-        limit=limit_param,
-        mofo_relevant=mofo_relevant_param,
+        end_time=end,
+        limit=limit,
+        after=after,
+        mofo_relevant=mofo_relevant,
     )
+    return get_bulk_contacts_by_timestamp_or_4xx(db=db, **bulk_request.dict())
```
bsieber-mozilla, line 636: Converted query_params into a `BulkRequestSchema` object with `@validators` to handle the defaulting and validation logic. 

_Distributes values to `get_bulk_contacts_by_timestamp_or_4xx` in expanded fashion: `**bulk_request.dict()`_

ctms/schemas/bulk.py
```diff
@@ -0,0 +1,54 @@
...
+
+class BulkRequestSchema(ComparableBase):
+    """A Bulk Read Request."""
+
+    start_time: datetime
+
+    end_time: Optional[Union[datetime, Literal[""]]] = None
+
+    @validator("end_time")
+    def end_time_must_not_be_blank(cls, value):  # pylint: disable=E0213
```
jwhitlock, line 20: E0213 is the warning "E0213: Method should have "self" as first argument (no-self-argument)". I have a preference for the string version, ``# pylint: disable=no-self-argument``, but it shorter and just a search away.

I checked if we're doing something wrong, but this seems to be a [bad interaction between pylint and pydantic being compiled with cython](https://github.com/samuelcolvin/pydantic/issues/568).

ctms/schemas/bulk.py
```diff
@@ -0,0 +1,54 @@
...
+        if value in BLANK_VALS:
+            return datetime.now(timezone.utc)
+        return value
+
+    limit: Optional[Union[int, Literal[""]]] = 10
```
jwhitlock, line 25: we should add more constraints to the int as well. It should be greater than 0, and have some upper limit like 100 or 1000. I think [conint](https://pydantic-docs.helpmanual.io/usage/types/#constrained-types) will do the heavy lifting. Maybe this?

```suggestion
    limit: Optional[Union[conint(gt=0, le=100), Literal[""]]] = 10
```

I've asked our partner for advice on the upper limit.

tests/unit/test_api.py
```diff
@@ -402,6 +402,11 @@ def test_get_ctms_for_api_example(client, example_contact):
         "/updates?start=2020-01-22T03%3A24%3A00%2B00%3A00&end=&limit=&after=&mofo_relevant=",
         None,
     ),
+    (
+        "GET",
+        "/updates?start=2020-01-22T03%3A24%3A00%2B00%3A00&end=&limit=&after=null&mofo_relevant=",
```
jwhitlock, line 407: As mentioned above, I think this should be a 422 instead of being successful. If we treat it as omitting ``after``, then it would restart the query sequence, and might look like our API is returning duplicate results.

ctms/schemas/bulk.py
```diff
@@ -0,0 +1,54 @@
...
+class BulkRequestSchema(ComparableBase):
+    """A Bulk Read Request."""
+
+    start_time: datetime
+
+    end_time: Optional[Union[datetime, Literal[""]]] = None
+
+    @validator("end_time")
+    def end_time_must_not_be_blank(cls, value):  # pylint: disable=E0213
+        if value in BLANK_VALS:
```
jwhitlock, line 21: After reading [Validate Always](https://pydantic-docs.helpmanual.io/usage/validators/#validate-always), I'm not sure if this is called when value is ``None``. We may only be looking for blank strings here.

jwhitlock (approved):
Thanks @bsieber-mozilla, this is an improvement to move validation into the schema and use more pydantic. I think it is mergable, with some nits and further investigations.

ctms/schemas/bulk.py
jwhitlock: Our partner regularly uses 100, agrees that 1000 would work as an upper limit

ctms/schemas/bulk.py
```diff
@@ -0,0 +1,54 @@
+import base64
+from datetime import datetime, timezone
+from typing import Literal, Optional, Union
+
+from pydantic import validator
+
+from .base import ComparableBase
+
+BLANK_VALS = [None, "", "null"]  # TODO: Should "null" be allowed?
```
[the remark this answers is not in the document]
bsieber-mozilla: Null should be removed now. Resolving.
https://github.com/mozilla-it/ctms-api/pull/208/files#diff-a30df6a18e8ac23b5a3a292ce386cbee84a11a3f462fd4690d7771dd683bed48R9

ctms/schemas/bulk.py
bsieber-mozilla: During testing I think None got here, because the fields were Optional?

ctms/schemas/bulk.py
bsieber-mozilla: Changed to the string version: resolving

ctms/schemas/bulk.py
jwhitlock: Perfect, thanks for confirming.

ctms/schemas/bulk.py
bsieber-mozilla: Moved default to 100, upper limit to 1000. Pushing in a new commit

tests/unit/test_api.py
bsieber-mozilla: "null" is now invalid

ctms/schemas/bulk.py
jwhitlock: I think 10 is still a good default, for someone just clicking buttons in the swagger docs 😬

jwhitlock:
Ug, looking at https://github.com/samuelcolvin/pydantic/issues/156, it appears that ``conint`` and ``mypy`` don't like each other at all. Sorry I steered you into the undocumented mess.

I think it would be best to go back to ``int``, an add checks in the validator ``limit_must_not_be_blank``, like:

```python
if value < 0:
    raise ValueError('"limit" should be greater than 0')
if value > 1000:
    raise ValueError('"limit"' should be less than or equal to 1000')
```

bsieber-mozilla:
> Ug, looking at [samuelcolvin/pydantic#156](https://github.com/samuelcolvin/pydantic/issues/156), it appears that `conint` and `mypy` don't like each other at all. Sorry I steered you into the undocumented mess.
> 
> I think it would be best to go back to `int`, an add checks in the validator `limit_must_not_be_blank`, like:
> 
> ```python
> if value < 0:
>     raise ValueError('"limit" should be greater than 0')
> if value > 1000:
>     raise ValueError('"limit"' should be less than or equal to 1000')
> ```

Made the changes after fighting with conint for a tiny bit; then my db went up in flames. I hit my limit for docker space. 🔥
_Did a rebase and forced the history for the branch_

Should be working and passing now.

ctms/schemas/bulk.py
```diff
@@ -0,0 +1,54 @@
...
+    @validator("after")
+    def after_must_be_base64_decodable(cls, value):  # pylint: disable=E0213
+        if value in BLANK_VALS:
+            return None
+        try:
+            str_decode = base64.urlsafe_b64decode(value)
+            str(str_decode.decode("utf-8")).split(
+                ","
+            )  # 'after' should be decodable otherwise err and invalid
+            return value
```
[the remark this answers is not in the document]
bsieber-mozilla: Returning the decoded "after" now. So no longer should required two-decoding attempts

PR #210, opened 2021-05-12
Add "trivial": true to logs for bots and boring requests

jwhitlock:
Move the tests for ``__heartbeat__`` and ``__lbheartbeat__`` from ``test_app.py`` into ``test_metrics.py``.

Add field ``"trivial": true`` for requests that we'd like to ignore when looking at logs:

* Requests to the root URL (redirected to ``/docs``)
* Amazon or New Relic requesting ``/__heartbeat__``
* Kubernetes requesting ``/__lbheartbeat__``
* Prometheus requesting ``/metrics``

This will allow us to write a Papertrail query that omits these noisy requests for issue #153, such as:

```
ctms-api-stage program:ctms -"\"trivial\": true"
```

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -396,6 +396,12 @@ async def log_request_middleware(request: Request, call_next):
         context = request.state.log_context
         if request.path_params:
             context["path_params"] = request.path_params
+        if "trivial_code" in context:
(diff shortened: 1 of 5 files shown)


imbstack (approved):
good call

PR #214, opened 2021-05-13
Rearrange tests

jwhitlock:
Move ``ctms/sample_data.py`` to ``tests/unit/sample_data.py``, since the samples are no longer used in the app itself and have been further tuned for tests.

To avoid exceeding 1000 lines in ``test_api.py``, split out into endpoint-specific tests:

* ``test_api_get.py`` - for ``GET /ctms/{email_id}``
* ``test_api_get_by_alt_id.py`` - for ``GET /ctms/?alt_id=id``
* ``test_api_post.py`` - for ``POST /ctms``
* ``test_api_put.py`` - for ``PUT /ctms/{email_id}``

Some cross-endpoint tests still remain in ``test_api.py``. The fixtures ``post_contact`` and ``put_contact`` have been brought over to ``conftest.py``.

I was tempted to try to combine some of the common functionality, but 1) they can't be merged entirely because they are both used in ``test_api::test_post_get_put``, and 2) it is not critical for launch, so I filed https://github.com/mozilla-it/ctms-api/issues/215. I avoided any other cleanup, so this should just be a re-arranging of test code without changes.


bsieber-mozilla (approved):
LGTM, thanks for breaking out the tests!

Issue #211, opened 2021-05-12
Add tracing via email string


jwhitlock:
For debugging, I'd like to see the request body for API calls for some users, while omitting them by default. My idea is to look for the string `+trace-me-` in the username portion of the primary email, such as `contact+trace-me-2021@gmail.com`. This allows adding a test email, using [Google's support for creating new email addresses](https://gmail.googleblog.com/2008/03/2-hidden-ways-to-get-more-from-your.html), in Basket or other systems, and then seeing the details of requests for that contact.

I also considered a tracing header, but that would require client changes, and I'm mostly interested in clients where I don't have control or visibility into their requests. It would be useful when debugging multi-system issues like #184.

Would `+trace-me-` be sufficient opt-in to allow logging of request details?

cvalaas:
I think it's a great idea!
The string needs to be unique enough that real users won't accidentally trigger it, which is what I guess you're asking. It _seems_ like `+trace-me-` would be enough, but I'm not positive.
I'd think for sure adding `+trace-me-mozilla-` would, though that's a mouthful (keyboardful?)

jwhitlock:
Yes, I'm looking for something that won't be accidental, so `+trace-me-mozilla-` is better. I'm not as worried as long as it is below 30 characters or so.

Maybe a random-looking passphrase is better? 1password suggests:
* `+trace-juggle-CAREFUL-`
* `+trace-oath-LETTER-`
* `+trace-thorned-DISORDER-`
* `+trace-gumdrop-PRANK-`
* `+trace-RANCID-swig-`

cvalaas:
I think `+trace-me-mozilla-` is more informative (and easier to remember), but I do love `+trace-gumdrop-PRANK-` and `+trace-RANCID-swig-`

jwhitlock:
Tested when debugging the issue of a Preference Center unsubscribe not setting the opt-out flag

Issue #173, opened 2021-04-30
Restructure Schemas


jwhitlock:
The current schemas almost work. The "PATCH" schemas seem like a better base schema (all optional fields), and switching would eliminate a current mypy error around the ``EmailPatchSchema`` and the ``EmailBase`` schema for the ``primary_email`` field (Incompatible types in assignment (expression has type "Optional[str]", base class "EmailBase" defined the type as "str"). One possible hierarchy:

* **Base schema** - all data items are optional, no DB-only fields like timestamps (appropriate for PATCH input, API methods)
  * **Create schema** - Some items are required, others have defaults if omitted (appropriate for POST input)
    * **Upsert schema** - More items are required (appropriate for PUT input)
      * **Output schema** - All items are required, includes timestamps (appropriate for API output)
        * **Model schema** - Adds DB-only fields, timestamp setters (appropriate for bulk importer)

In the case of emails:
* ``EmailBaseSchema`` - All data fields optional
  * ``EmailCreateSchema`` - ``primary_email`` now required, view will generate ``email_id`` if omitted
    * ``EmailUpsertSchema`` - ``email_id`` now required
      * ``EmailOutputSchema`` - All data is required, adds timestamps, ``orm_mode=True``
        * ``EmailModelSchema`` - Adds `extra="forbid"` to forbid extra fields. In the future may include additional DB-only columns.

I'd also like to:
* Use ``class Config`` for descriptions and examples. I'm jealous of Taskcluster's structured data for API definitions, and I see this as a transitional step, as well as removing some duplication in the docs
* Make mypy happier, maybe even required for schemas
* Make the output of full contact data more natural. There's too much manual creation of ``ContactSchema`` and ``amo=contact.amo or AddOnsSchema()`` nonsense.

This all seems possible to me, but I expect details to change as I restructure the code.

PR #216, opened 2021-05-13
Add log tracing for emails with "+trace-me-mozilla-"

jwhitlock:
As suggested in issue #211, add a feature that traces requests if contact's primary email has the string ``+trace-me-mozilla-``, such as ``test+trace-me-mozilla-1@example.com``. The contact itself is examined, so the string does not need to appear in the request body.

Tracing is enabled for these API endpoints:

* ``GET /ctms``
* ``POST /ctms``
* ``GET /ctms/{email_id}``
* ``PUT /ctms/{email_id}``
* ``PATCH /ctms/{email_id}``

When active, two items are added to the log:

* ``trace={email}``, placing the email in the log. For ``GET /ctms``, if multiple emails are traced, the value is a comma-separated list of the traced emails
* ``trace_json={request_json}``, when the request has a JSON body (``POST``, ``PUT``, ``PATCH``)

The purpose is to allow debugging requests from an integrated service, such as Basket or Mozilla Foundation systems. The string was chosen for the very small chance of adding it by accident, and because of the implied consent to tracing implied by the text.

~~This is built on PR #214, and adds two more commits. I can rebase to make reviewing this one easier.~~ Merged and rebased

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -436,6 +436,7 @@ def root(request: Request):
     tags=["Public"],
 )
 def read_ctms_by_any_id(
+    request: Request,
     db: Session = Depends(get_db),
     api_client: ApiClientSchema = Depends(get_enabled_api_client),
     ids=Depends(all_ids),
@@ -446,6 +447,13 @@ def read_ctms_by_any_id(
         )
         raise HTTPException(status_code=400, detail=detail)
     contacts = get_contacts_by_ids(db, **ids)
+    traced = []
+    for contact in contacts:
+        email = contact.email.primary_email
+        if re_trace_email.match(email):
+            traced.append(email)
+    if traced:
+        request.state.log_context["trace"] = ",".join(traced)
     return [
         CTMSResponse(
             amo=contact.amo or AddOnsSchema(),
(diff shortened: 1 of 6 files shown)


bsieber-mozilla (approved):
LGTM!

_I wonder if adding this to the middleware of the app would make any sense; with the variety of Response types, it could be annoying to filter through for their emails... It looks good as is, so we could make a PR later if desired._

cvalaas (approved):
lgtm

jwhitlock:
@bsieber-mozilla asked:

> I wonder if adding this to the middleware of the app would make any sense; with the variety of Response types, it could be annoying to filter through for their emails... It looks good as is, so we could make a PR later if desired.

I tried this as my first approach. The issue is that FastAPI assumes it is the first to look at the request, and consumes the entire request, so you can't read the request content in the middleware with ``await request.json()``. If the middleware reads the request before FastAPI, then FastAPI awaits zero remaining bytes forever, and if the middleware tries to read afterwards, then it awaits forever.

It may be possible to extend FastAPI so that the request body is stored on the request, and then at least that part could be done in the middleware. I didn't want to get into that with a deadline.

Once my dreams of using the middleware were over, I just dug in and changed all the endpoints. I suspect there might be similar issues trying to read the response in the middleware - if the middleware reads it, then nothing might be returned to the external caller.

I think it is worth trying again to centralize the code, but I wouldn't tackle it myself until after issue #173, which will be after launch at this rate.

PR #218, opened 2021-05-17
Setup multiprocessing metrics registry in ``/metrics`` endpoint

jwhitlock:
For each process, use a separate ``CollectorRegistry`` initialized at startup.

For the ``/metrics`` view, use this collecter in single-process mode, but use a fresh ``CollectorRegistry`` with 
``MultiProcessCollector`` in multiprocess under ``gunicorn``. This should avoid duplicate metric exports (one for multiprocess, one for the process serving the request).

This was based off of reading [starlette_explorer](https://github.com/stephenhillier/starlette_exporter)'s middleware and view, as well as finally understanding this line of the [multiprocessing docs](https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn):

> It is a best practice to create this registry inside the context of a request to avoid metrics registering themselves to a collector used by a ``MultiProcessCollector``. If a registry with metrics registered is used by a ``MultiProcessCollector`` duplicate metrics may be exported, one for multiprocess, and one for the process serving the request.

This suggests that in the stage metrics, the requests counts jump around because the metrics are the group metrics plus those collected by that process.

CTMS jumping metrics

I verified the issue and the solution with a temporary PID-based metric, see https://github.com/mozilla-it/ctms-api/commit/310228408c6d27f2becba8adab7de70695448315 for code.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -11,7 +11,7 @@ import uvicorn
 from fastapi import Depends, FastAPI, HTTPException, Path, Request, Response
 from fastapi.responses import RedirectResponse
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
-from prometheus_client import CONTENT_TYPE_LATEST, generate_latest
+from prometheus_client import CONTENT_TYPE_LATEST, CollectorRegistry, generate_latest
 from pydantic import ValidationError
 from sentry_sdk.integrations.asgi import SentryAsgiMiddleware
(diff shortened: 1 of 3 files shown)


imbstack (approved):
makes sense to me

Thanks for this, looks like it took some good digging to figure out

bsieber-mozilla (approved):

PR #209, opened 2021-05-12
Adding a docker-compose lint check, that uses the dockerfile's multistage lint target

bsieber-mozilla:
A potential solution for : https://github.com/mozilla-it/ctms-api/issues/66

adds:
`make lint`

_which rather checks to confirm/deny if linting has been done; rather than to lint the code._

---

💭 Should this be `lint-check`? 
💭 Would we want a `make lint` to lint the code as well? (Is this preferred to pre-commit?)
🤔

Issue #66, opened 2021-02-25
Add "make lint"


jwhitlock:
Add a `make lint` command that runs the linter inside the Docker container. 

Currently, ``script/lint.sh`` can only run inside a ``poetry shell`` on the development system, because ``detect-secrets-hook`` requires ``git``, which is not in the Docker container, after PR #12. I don't think we should add ``git`` to the container, since I don't think we should commit inside of the container either. Some options are to skip checking secrets when ``git`` is not available, or generate the file list without ``git``.

bsieber-mozilla:
A potential solution could be this...https://github.com/mozilla-it/ctms-api/issues/66

`make lint` should do the lint checking through docker

jwhitlock:
As of the recent changes, we have these linters:

* [bandit](https://github.com/PyCQA/bandit) - Security analysis of Python code
* [black](https://black.readthedocs.io/en/stable/) - Formats Python code
* [detect-secrets](https://github.com/Yelp/detect-secrets) - Look for secrets in code an docs
* [isort](https://github.com/PyCQA/isort) - Sorts Python imports
* [seed-isort-config](https://github.com/asottile-archive/seed-isort-config) - Statically populate the ``known_third_party`` ``isort`` setting. (deprecated with `isort` 5)
* [mypy](https://mypy.readthedocs.io/en/stable/) -Static type checker for Python
* [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) - A collection of hooks for [pre-commit](https://github.com/pre-commit/pre-commit), including:
  - [check-added-large-files](https://github.com/pre-commit/pre-commit-hooks#check-added-large-files) - Prevent committing giant files
  - [check-json](https://github.com/pre-commit/pre-commit-hooks#check-json) - Attempts to load all json files to verify syntax.
  - [detect-private-key-file](https://github.com/pre-commit/pre-commit-hooks#detect-private-key)
  - [end-of-file-fixer](https://github.com/pre-commit/pre-commit-hooks#end-of-file-fixer) - Makes sure files end in a newline and only a newline
  - [pretty-format-json](https://github.com/pre-commit/pre-commit-hooks#pretty-format-json) - Checks that all your JSON files are pretty
  - [trailing-whitespace](https://github.com/pre-commit/pre-commit-hooks#trailing-whitespace) - Trims trailing whitespace. 
* [pylint](http://pylint.pycqa.org/en/latest/) - Enforces Python coding standards

[...]

PR #222, opened 2021-05-19
More linting adjustments

jwhitlock:
Update the lint target, which runs with `make lint` and as a build step in CirceCI, to match `scripts/lint.sh` a little closer. The order of tests have changed, `mypy` uses `mypy.ini` instead of command line switches, and `pylint` is added.

``docker-compose.lint.yml`` now lives in the ``test`` folder.

``scripts/lint.sh`` now checks for ``git`` before running ``detect-secrets-hook``. In general, ``detect-secrets`` seems to work best as a ``pre-commit`` hook. I'm still not sure how to run it as a CI step, but running it in-container seems like a dead end.

This is for issue #66, continuing @bsieber-mozilla's work in PR #209.


bsieber-mozilla (approved):
This looks like a good set of changes to the `lint` target. I think up to this point, the script was used as an alternate path for linting instead of through pre-commit, and the lint target was for confirming that the code had been linted; we can change the behavior to reduce redundancy and leave pre-commit as the main lint and secret checker (maybe in another PR?)

imbstack (approved):

jwhitlock:
> I think up to this point, the script was used as an alternate path for linting instead of through pre-commit, and the lint target was for confirming that the code had been linted; we can change the behavior to reduce redundancy and leave pre-commit as the main lint and secret checker (maybe in another PR?)

I'll need to think on that a bit. I still want to run `black` and `pylint` and `mypy` outside of a commit, and `scripts/lint.sh` is a quick way to do that, `make lint` will tell me when I'm wrong but not fix stuff like `black`. But, I may be the only dev who does that, and I should get used to running individual commands or manually running `pre-commit`. Let me collect my thoughts and put them on issue #66

PR #225, opened 2021-05-19
Initial commit for enabling case-insensitivity, calling 'lower()' on primary_email's.

bsieber-mozilla:
Previous solution: Use @validator decorators to convert incoming values to lowercase, so they are lowercase in the DB.

New solution: Don't modify default or validation logic, just add comparators for LC comparison on DB

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -198,7 +198,9 @@ def get_emails_by_any_id(
     if email_id is not None:
         statement = statement.filter(Email.email_id == email_id)
     if primary_email is not None:
-        statement = statement.filter(Email.primary_email == primary_email)
+        statement = statement.filter_by(
+            primary_email_insensitive_comparator=primary_email
+        )
     if basket_token is not None:
         statement = statement.filter(Email.basket_token == str(basket_token))
     if sfdc_id is not None:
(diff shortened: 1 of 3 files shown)


bsieber-mozilla:
A brief draft after following along on slack.

jwhitlock:
I think we just want a case-insensitive comparison, not enforce that all emails are lowercase. See https://stackoverflow.com/questions/53063778/case-insensitive-indexing-in-postgres-with-sqlalchemy

bsieber-mozilla:
Manual upgrade and rollback migration required with functional indexes. Comments from alembic in-line to migration script.

bsieber-mozilla:
# Indexes
Screen Shot 2021-05-19 at 5 38 52 PM
Screen Shot 2021-05-19 at 5 39 01 PM

bsieber-mozilla:
# Case Insensitive Reading

Screen Shot 2021-05-19 at 5 47 33 PM
Screen Shot 2021-05-19 at 5 47 04 PM

bsieber-mozilla:
## POST: of Contact Creation used above
Screen Shot 2021-05-19 at 5 46 47 PM

bsieber-mozilla:
@jwhitlock I went through the stack overflow and made changes to just the indexing and querying at this time. Models should no longer augment the data/email casing when put into the db.

jwhitlock (approved):
Thanks, this looks good, and your screenshots look like it is working. Good job including the `fxa_primary_email`!

I'd like to verify that Postgres is using the index for these lookups. I don't have a well-populated dev database, so I think it is easiest on stage.

bsieber-mozilla:
Issues to be created: 
- Make FXA unique?
- Remove unnecessary indexes

bsieber-mozilla:
Going to merge this in a few moments, unless someone stops me

PR #227, opened 2021-05-20
Modifying migration

bsieber-mozilla:
> alembic current reports cad3c933830b (head)

Since the prior migration failed and RB'd to the previous revision, this PR removes the uniqueness element to the func.lower(primary_email) index


jwhitlock (approved):
I think this will work, since the alembic revision hasn't been fully applied in stage.

You may want to manually downgrade /upgrade on your dev box, to get your DB consistent.

Issue #125, opened 2021-04-01
Add metrics to background jobs


jwhitlock:
Add metrics to background jobs, to make it more visible when they run and how long they take. The SRE team recommends a Prometheus middleware and exporter, which can be fed into our existing metrics infrastructure, including dashboards and metrics.

* [Instrumenting Python for Production](https://www.polibyte.com/presentations/instrumenting_python_for_production) presentation
* [Bedrock implementation](https://github.com/mozilla/bedrock/pull/8978) (Django) and [output](https://www.mozilla.org/prometheus/metrics)

Possible libraries:
* https://github.com/prometheus/client_python

Tracked in Jira:SE-1478, and as part of Jira:OSS-339 (maybe)

jwhitlock:
From https://github.com/mozilla-it/ctms-api/issues/124#issuecomment-834555033

> What would be great to see in the next phase is instrumentation around any outgoing requests you have. I know the webapp talks to postgres, I'm less sure if it talks to other systems like Acoustic directly. For anything it does talk to, having a counter for the total requests and a histogram of response time, both broken down by success vs error, would be helpful for debugging.

This seems like it should be in the initial metrics implementation for monitoring the background Acoustic job. My first pass at metrics:

* ``acoustic_request_total`` - Counter of acoustic requests. Labels are ``type`` or ``path`` (to distinguish between multiple types of requests) and ``status`` / ``status_code`` (to distinguish success and failures)
* ``acoustic_request_duration`` - Histogram of acoustic request timing. Labels are ``type`` / ``path``, and ``status`` / ``status_code``
* ``acoustic_sync_total`` - Counter of contacts synced to acoustic. Not sure if labels are needed, this may just count success
* ``acoustic_sync_backlog`` - Gauge of the number of contacts in the sync backlog. I don't think the background job will run under [multiprocess mode](https://github.com/prometheus/client_python#multiprocess-mode-gunicorn), but if it does, `max` seems like the right aggregator.

bsieber-mozilla:
`Last success`

leplatrem:
Background jobs were removed. Closing

PR #234, opened 2021-05-25
Adding Background Metrics

bsieber-mozilla:
An initial attempt to tackle: https://github.com/mozilla-it/ctms-api/issues/125

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -294,57 +294,50 @@ class CTMSToAcousticService:
             "sync_fields": sync_fields,
             "columns": columns,
         }
-        if self.metric_service:
+        if self.metric_service is None:
+            self.acoustic.add_recipient(**params)
+        else:  # Metrics are enabled
             start_time = time.monotonic()
-            err = None
             status = "success"
             try:
                 self.acoustic.add_recipient(**params)  # Call to Acoustic
-            except Exception as e:  # pylint: disable=broad-except
-                err = e
+            except Exception:  # pylint: disable=broad-except
                 status = "failure"
-            duration = time.monotonic() - start_time
-            duration_s = round(duration, 3)
-            metric_params = {"method": "add_recipient", "status": status}
-            self.metric_service.inc_acoustic_request_total(**metric_params)
-            metric_params.update({"duration_s": duration_s})
-            self.metric_service.observe_acoustic_request_duration(**metric_params)
-            if err:
-                # Delayed raise for metric analysis
-                raise err
-        else:
-            self.acoustic.add_recipient(**params)
+                raise
+            finally:
+                duration = time.monotonic() - start_time
+                duration_s = round(duration, 3)
+                metric_params = {"method": "add_recipient", "status": status}
+                self.metric_service.inc_acoustic_request_total(**metric_params)
+                metric_params.update({"duration_s": duration_s})
+                self.metric_service.observe_acoustic_request_duration(**metric_params)
 
     def _insert_update_newsletters(self, table_id=None, rows=None):
         if rows is None:
             rows = []
         params = {"table_id": table_id, "rows": rows}
-        if self.metric_service:
+        if self.metric_service is None:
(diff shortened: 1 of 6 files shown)


ctms/acoustic_service.py
```diff
@@ -236,11 +240,11 @@ def _newsletter_converter(self, acoustic_main_table, contact):
             if newsletter.name in AcousticResources.MAIN_TABLE_SUBSCR_FLAGS.keys():
                 newsletter_dict = newsletter.dict()
                 _today = datetime.date.today().isoformat()
-                newsletter_template["create_timestamp"] = self.dictionary_helper(
-                    newsletter_dict, "create_timestamp", _today
+                newsletter_template["create_timestamp"] = newsletter_dict.get(
```
jwhitlock, line 243: Thank you!

ctms/acoustic_service.py
```diff
@@ -288,22 +286,65 @@ def _add_contact(self, list_id=None, sync_fields=None, columns=None):
...
+        if self.metric_service:
+            start_time = time.monotonic()
+            err = None
+            status = "success"
+            try:
+                self.acoustic.add_recipient(**params)  # Call to Acoustic
+            except Exception as e:  # pylint: disable=broad-except
+                err = e
+                status = "failure"
+            duration = time.monotonic() - start_time
```
jwhitlock, line 306: I think the metrics stuff should be in a ``finally`` clause, so that exceptions raised by the metrics code doesn't hide exceptions from acoustic. Something like:

```python
except Exception as e:
  status = "failure"
  raise
finally:
  duration = time.monotonic() - start_time
  # The rest of the metrics code
``` 

I'm tempted to suggest a metrics context manager, but I think that would obfuscate otherwise straightforward code.

jwhitlock (commented):
Thanks @bsieber-mozilla, looking good so far!

duallain:
Is it easy to get a sample of the metrics that will be emitted? Interested in what names and tags we end up with. The code all looks reasonable to me on first pass, but hard to see what we're going to end up with it from this.

bsieber-mozilla:
Screen Shot 2021-05-26 at 10 04 25 AM
Screen Shot 2021-05-26 at 10 04 15 AM
Screen Shot 2021-05-26 at 10 04 03 AM

duallain:
Ahh, thank you @bsieber-mozilla - couple of small comments:

All the existing metrics start with ctms_, do you think of these metrics as being a subset of ctms? It may make sense to call it ctms_acoustic_request_created if you do. That will make it easier to find and tie them together.  We're pushing to a web-sre db in influx, and at the moment only ctms is in that db, but soon I imagine we'll add more, and then things will get muddied I assume.

I think the component label is wrong. web would indicate the metric comes from the web pod, but this is coming from what I called 'background'.

For additional metrics, Count of failures and count of retries has also been helpful for basket. Being able to see a spike in failures for a particular point in time, or a huge amount of retries can help explain why the queue is backing up.

All that being said, this does look good to me.  I think I'd ask for the component label to be changed, but the other two suggestions could be skipped for now, and we merge as is (up to you).

duallain (approved):
FWIW, all my comments so far are ok to be handled later, I think merging this in before we get another huge batch of requests would be great.  This looks good enough to merge as is, if you're happy with it.

ctms/acoustic_service.py
bsieber-mozilla: I appreciated the call out, I was hitting my head; almost made a default_dict and ended up just recreating the extended `get` without realizing it. Using get with a default value...well I'm really happy that you corrected me so I can do this appropriately

bsieber-mozilla:
Added a prefix to the metrics of `ctms_background_`

bsieber-mozilla:
I say lets try this, and while I work through some other acoustic changes, I can modify any of these paths that may not work as expected when live.

Merging soon, with some following PRs for acoustic changes to be coming today/tomorrow (regarding source change)

PR #235, opened 2021-05-26
Hf batch limit and commit more

bsieber-mozilla:
Introduce env var CTMS_ACOUSTIC_BATCH_LIMIT default to 20.

Commit after each record is processed

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -3,7 +3,7 @@ from datetime import datetime, timezone
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type, cast
 
 from pydantic import UUID4
-from sqlalchemy import asc, desc, or_
+from sqlalchemy import asc, or_
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session, joinedload, load_only, selectinload
 
@@ -274,29 +274,42 @@ def get_all_acoustic_records_before(
 ) -> List[PendingAcousticRecord]:
     """
     Get all the pending records before a given date. Allows retry limit to be provided at query time."""
-    if batch_limit is None:
-        pending_records: List[PendingAcousticRecord] = (
-            db.query(PendingAcousticRecord)
-            .filter(
-                PendingAcousticRecord.update_timestamp < end_time,
-                PendingAcousticRecord.retry < retry_limit,
-            )
-            .order_by(desc(PendingAcousticRecord.update_timestamp))
-            .all()
-        )
-        return pending_records
-    # else
-    batch_pending_records: List[PendingAcousticRecord] = (
+    query = (
         db.query(PendingAcousticRecord)
         .filter(
             PendingAcousticRecord.update_timestamp < end_time,
             PendingAcousticRecord.retry < retry_limit,
         )
-        .order_by(desc(PendingAcousticRecord.update_timestamp))
-        .limit(batch_limit)
-        .all()
+        .order_by(
+            asc(PendingAcousticRecord.update_timestamp),
+            asc(PendingAcousticRecord.retry),
+        )
     )
-    return batch_pending_records
+    if batch_limit:
+        query = query.limit(batch_limit)
+
+    pending_records: List[PendingAcousticRecord] = query.all()
+    return pending_records
+
+
+def get_all_acoustic_records_count(
+    db: Session, end_time: datetime, retry_limit: int = 5
+) -> int:
+    """
+    Get all the pending records before a given date. Allows retry limit to be provided at query time."""
+    query = (
+        db.query(PendingAcousticRecord)
+        .filter(
+            PendingAcousticRecord.update_timestamp < end_time,
+            PendingAcousticRecord.retry < retry_limit,
+        )
+        .order_by(
+            asc(PendingAcousticRecord.update_timestamp),
+            asc(PendingAcousticRecord.retry),
+        )
+    )
+    pending_records_count: int = query.count()
+    return pending_records_count
 
 
 def get_acoustic_record_as_contact(
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -269,25 +269,28 @@ def get_contacts_by_any_id(
     return data
 
 
-def get_all_acoustic_records_before(
-    db: Session, end_time: datetime, retry_limit: int = 5, batch_limit=None
-) -> List[PendingAcousticRecord]:
-    """
-    Get all the pending records before a given date. Allows retry limit to be provided at query time."""
-    query = (
+def _acoustic_sync_base_query(db: Session, end_time: datetime, retry_limit: int = 5):
+    return (
         db.query(PendingAcousticRecord)
         .filter(
             PendingAcousticRecord.update_timestamp < end_time,
             PendingAcousticRecord.retry < retry_limit,
         )
         .order_by(
-            asc(PendingAcousticRecord.update_timestamp),
             asc(PendingAcousticRecord.retry),
+            asc(PendingAcousticRecord.update_timestamp),
         )
     )
+
+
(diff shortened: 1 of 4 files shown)


cvalaas (approved):

bsieber-mozilla:
Sorry for the (un+)assign Alan, meant to tag as review for awareness and insight.

ctms/crud.py
```diff
@@ -270,23 +270,33 @@ def get_contacts_by_any_id(
...
     Get all the pending records before a given date. Allows retry limit to be provided at query time."""
-    pending_records: List[PendingAcousticRecord] = (
+    if batch_limit is None:
+        pending_records: List[PendingAcousticRecord] = (
+            db.query(PendingAcousticRecord)
+            .filter(
+                PendingAcousticRecord.update_timestamp < end_time,
+                PendingAcousticRecord.retry < retry_limit,
+            )
+            .order_by(asc(PendingAcousticRecord.update_timestamp))
```
cvalaas, line 284: maybe switch this to `desc` for now

bsieber-mozilla: changed this

bsieber-mozilla: With these changes...we could also just give it limit 1 as envvar and spin multiple up maybe without stepping on toes too much...? 

There might be more to do to make it more parallelizable though...

cvalaas: i  think limit 1  is  too few.  to actually make it parallelizable we need a lock or  partition or somesuch

ctms/sync.py
```diff
@@ -90,12 +92,15 @@ def sync_records(self, db, end_time=None):
...
+            retry_limit=self.retry_limit,
+            batch_limit=self.batch_limit,
         )
 
         # For each record, attempt downstream sync
         for acoustic_record in all_acoustic_records_before_now:
             self._sync_pending_record(db, acoustic_record)
-        # Commit changes to db after all records are processed
-        db.commit()
+            # Commit changes to db after EACH record is processed
```
cvalaas, line 104: ironically, if we're only doing 20 at a time, this probably  isn't necessary :) 🤷

duallain:
One suggestion, include number of retries as part of the ordering.  I think that transactions that have never failed before are more likely to succeed, so if the goal is to keep fewest items in flight that would be a good approach.  If the goal is to have lowest worst case latency, preferring retries > 1 might be good.  (May have missed this, but I think you're just ordering by update timestamp at the moment).

ctms/crud.py
```diff
@@ -270,23 +270,33 @@ def get_contacts_by_any_id(
...
+        return pending_records
+    # else
+    batch_pending_records: List[PendingAcousticRecord] = (
         db.query(PendingAcousticRecord)
         .filter(
             PendingAcousticRecord.update_timestamp < end_time,
             PendingAcousticRecord.retry < retry_limit,
         )
-        .order_by(asc(PendingAcousticRecord.update_timestamp))
+        .order_by(desc(PendingAcousticRecord.update_timestamp))
```
jwhitlock, line 295: Is the switch to ``desc`` (last in first out) a permanent thing, or do we want to switch back to ``asc`` (first in first out) eventually?

ctms/crud.py
```diff
@@ -270,23 +270,33 @@ def get_contacts_by_any_id(
...
 def get_all_acoustic_records_before(
-    db: Session,
-    end_time: datetime,
-    retry_limit: int = 5,
+    db: Session, end_time: datetime, retry_limit: int = 5, batch_limit=None
 ) -> List[PendingAcousticRecord]:
     """
     Get all the pending records before a given date. Allows retry limit to be provided at query time."""
-    pending_records: List[PendingAcousticRecord] = (
+    if batch_limit is None:
```
jwhitlock, line 277: I think it would be clearer that the queries are similar if we move the ``batch_limit`` logic down, something like:

```python
pending_records: List[PendingAcousticRecord] = (
   db.query(PendingAcousticRecord)
   ...
)
if batch_limit:
  pending_records = pending_records.limit(batch_limit)
return pending_records.all()
```

ctms/crud.py
```diff
@@ -270,23 +270,33 @@ def get_contacts_by_any_id(
...
+        pending_records: List[PendingAcousticRecord] = (
+            db.query(PendingAcousticRecord)
+            .filter(
+                PendingAcousticRecord.update_timestamp < end_time,
+                PendingAcousticRecord.retry < retry_limit,
+            )
+            .order_by(desc(PendingAcousticRecord.update_timestamp))
+            .all()
```
jwhitlock, line 285: We may want to add a [row lock](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.with_for_update), like:

```
.with_for_update()
.all()
```

jwhitlock (approved):
Thanks for the quick addition of batching, I think batching makes sense as the default moving forward. Let's get something running and work out the bugs.

ctms/crud.py
jwhitlock: I see from PR #234 that we're also using this function to count the size of the backlog, which means we shouldn't unconditionally lock for updates. This suggests it should be another parameter, or even better, split into a crud function to count the backlog size vs get a batch of records.

ctms/sync.py
bsieber-mozilla: Reverted

bsieber-mozilla:
> One suggestion, include number of retries as part of the ordering. I think that transactions that have never failed before are more likely to succeed, so if the goal is to keep fewest items in flight that would be a good approach. If the goal is to have lowest worst case latency, preferring retries > 1 might be good. (May have missed this, but I think you're just ordering by update timestamp at the moment).

Added to the ordering the retry logic here: https://github.com/mozilla-it/ctms-api/pull/235/commits/a74b6d6cf17c1d3cbb9388cdd7c49008f20d855c#diff-45799bd240716e4cd04a9c1aeb24ad5e995dd9587db55a53d9eb85aa182d3533R285

I think this makes sense. 

Although, upon a "retry" I believe we are updating the update_timestamp, which would seemingly put the failed record back as the last entered record, until a new record was scheduled. 

This does make it so records that have err'd before wait until all other records are processed. 

I'm on the fence with it; if we get into any failures that retry would be success for those would be delayed. (But if they failed once... it's likely they'll just keep failing.)

ctms/crud.py
bsieber-mozilla: Did the needful, changed to reduced duplication; thanks for the suggestion

ctms/crud.py
bsieber-mozilla: I think it was a temporary thing, given the lack of observability of the system and to unblock testing. I've reverted to asc.

ctms/crud.py
```diff
@@ -274,29 +274,42 @@ def get_all_acoustic_records_before(
...
-    batch_pending_records: List[PendingAcousticRecord] = (
+    )
+    if batch_limit:
+        query = query.limit(batch_limit)
+
+    pending_records: List[PendingAcousticRecord] = query.all()
+    return pending_records
+
+
+def get_all_acoustic_records_count(
```
bsieber-mozilla, line 295: split this out for pr #234

ctms/crud.py
bsieber-mozilla: I did not add the row lock, but can still do so; for PR #234 I'll need to go back and rebase on these changes, but I added a separate method to count. _(Still can improve with making a _bg_base_query that can be reused by them both)_

jwhitlock: let's add the row lock later, we should make sure we understand how to release the lock after the update, rather than leave open forever transactions...

bsieber-mozilla:
Will merge shortly... @duallain, I flipped the query filters

PR #237, opened 2021-06-01
import updates 2

imbstack:


scripts/importers/finish.sql
```diff
@@ -12,6 +11,12 @@ alter table newsletters add constraint uix_email_name unique (email_id, "name");
 alter table vpn_waitlist add constraint vpn_waitlist_email_id_key unique (email_id);
 alter table fxa add constraint fxa_fxa_id_key unique (fxa_id);
 
+-- Easy enough to add this now too
+create index bulk_read_index on emails (update_timestamp, email_id);
+
+-- New matching
+CREATE UNIQUE INDEX idx_email_primary_unique_email_lower ON public.emails USING btree (lower((primary_email)::text));
```
jwhitlock, line 18: Thanks for the reminder, I need to do some model and index cleanup. Filed https://github.com/mozilla-it/ctms-api/issues/238.

scripts/importers/finish.sql
```diff
@@ -44,3 +49,4 @@ alter table mofo add constraint mofo_email_id_fkey foreign key (email_id) refere
 
 -- One more cleanup
 update newsletters set "format" = 'T' where "format" = 'N';
+alter table newsletters drop column source, add column source text;
```
jwhitlock, line 52: What was the purpose of this cleanup?

jwhitlock (approved):
Thanks @imbstack!

scripts/importers/finish.sql
imbstack: That was the empty-string for source instead of null cleanup

PR #243, opened 2021-06-25
Adding conversion into C->A backend sync for fxa_created_date

bsieber-mozilla:
Relevant:
https://docs.google.com/document/d/10UPt9kSE0ynoBK-Tv5JOufkGdMwMogdmAbEoLc_apkk/edit

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -5,6 +5,7 @@ from decimal import Decimal
 from typing import List
 from uuid import UUID
 
+import dateutil
 from lxml import etree
 from silverpop.api import Silverpop, SilverpopResponseException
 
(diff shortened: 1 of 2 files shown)


ctms/acoustic_service.py
```diff
@@ -212,6 +213,11 @@ def _main_table_converter(self, contact):
                         acoustic_field_name
                         in AcousticResources.VALID_ACOUSTIC_MAIN_TABLE_FIELDS
                     ):
+                        if acoustic_field_name == "fxa_created_date":
+                            inner_value = self.fxa_created_date_string_to_datetime(
+                                inner_value
+                            )
+
```
bsieber-mozilla, line 220: If fxa_created_date is a string it will be parsed into a datetime, if it was not an instanceof string then it's a no-op.

The inner_value is converted from datetime to acoustic_readable in the following few lines `self.transform_field_for_acoustic`

cvalaas (approved):

PR #244, opened 2021-06-25
Updating from ISO to MM/DD/YYYY

bsieber-mozilla:
Related:
> To follow-up tomorrow: possibly none of the dates we're sending from C->A are in a format that Acoustic likes. We're sending YYYY-MM-DD but Acoustic might need MM/DD/YYYY (ref: https://help.goacoustic.com/hc/en-us/articles/360043349433-Work-with-database-fields)
> Should be a(nother) simple change to the C->A script
> ...https://mozilla.slack.com/archives/G01BEQJFWS3/p1624593115185800

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -299,9 +299,9 @@ class CTMSToAcousticService:
             return "0"
         if isinstance(data, datetime.datetime):
             # Acoustic doesn't have timestamps, so make timestamps into dates.
-            return data.date().isoformat()
+            data = data.date()
         if isinstance(data, datetime.date):
-            return data.isoformat()
+            return data.strftime("%m/%d/%Y")
         if isinstance(data, UUID):
             return str(data)
         return data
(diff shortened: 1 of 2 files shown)


ctms/acoustic_service.py
```diff
@@ -299,9 +299,9 @@ def transform_field_for_acoustic(data):
             return "0"
         if isinstance(data, datetime.datetime):
             # Acoustic doesn't have timestamps, so make timestamps into dates.
-            return data.date().isoformat()
+            data = data.date()
         if isinstance(data, datetime.date):
-            return data.isoformat()
+            return data.strftime("%m/%d/%Y")
```
bsieber-mozilla, line 304: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes



%m | Month as a zero-padded decimal number. | 01, 02, …, 12 | (9)
-- | -- | -- | --




%d | Day of the month as a zero-padded decimal number. | 01, 02, …, 31 | (9)
-- | -- | -- | --




%Y | Year with century as a decimal number. | 0001, 0002, …, 2013, 2014, …, 9998, 9999 | (2)
-- | -- | -- | --

ctms/acoustic_service.py
```diff
@@ -299,9 +299,9 @@ def transform_field_for_acoustic(data):
             return "0"
         if isinstance(data, datetime.datetime):
             # Acoustic doesn't have timestamps, so make timestamps into dates.
-            return data.date().isoformat()
+            data = data.date()
```
bsieber-mozilla, line 302: _...it does roll into the following `if` because it is a `date`_

cvalaas (approved):

PR #303, opened 2021-11-25
Test and fix stripe relationships

jwhitlock:
Relationships with out a ``ForeignKey`` are tricky. SQLAlchemy will fail to load if some configurations are used, but in others it generates a warning:

```
SAWarning: Multiple rows returned with uselist=False for lazily-loaded attribute (Model.attribute)
```

In this case, the SQL generated has an overly broad `WHERE` clause, and ALL rows with a match between the two tables are returned. SQLAlchemy expects one, so it emits the warning and picks one, probably the first. This appears to be the cause of an issue in stage, where the wrong `email_id` is scheduled for syncing when ingesting a new Stripe object for an existing user.

The ``SAWarning`` is only generated for one-to-one relationships, when there are multiple matches. A lot of our tests create a single contact, so we missed this. For one-to-many relationships, the same issue can cause all objects to be returned, rather than just the ones with the matching IDs.

This PR expands the tests with a fixture with doubles of everything:

* Two contacts with FxA and a Stripe customer
* Two subscriptions per customer
* Two subscription items per subscription (4 per customer)
* Two invoices per customer
* Two invoice line items per invoice (4 per customer, linked to the 2 subscriptions and 4 subscription items)
* Four prices (one per subscription item, and both customers are subscribed to all four)

It then tests all the model relationships to and from Stripe models. This identified several errors, and several ``SAWarning``s, which are upgraded to errors for CRUD tests.

Some bizarre trial and error eliminated the errors. I don't feel like the SQLAlchemy docs really address this use case, of relations between models that do not involve a foreign key. Beware borrowing this code. However, all tests pass, with no ``SAWarning``s.

It was useful to see the generated SQL, to see the overly broad queries. This can be done with ``pytest -s -vvv``, but there is a ton of output so it is useful to limit to a single test, like ``pytest -sx -vvv tests/unit/test_crud.py::test_relations_to_stripe_objects[0]``.

It was also useful to see these when interacting with the website, ``POST``ing some Stripe data and seeing the results. This requires adding a logging handler for ``sqlalchemy.engine``. I added this as well, but disabled by default, since we don't want to log SQL in deployments.

### DIFF
--- a/ctms/models.py
+++ b/ctms/models.py
@@ -484,11 +484,10 @@ class StripeSubscription(StripeBase):
 
     customer = relationship(
         "StripeCustomer",
-        back_populates="subscriptions",
         uselist=False,
         primaryjoin=(
-            "foreign(StripeCustomer.stripe_id)=="
-            " remote(StripeSubscription.stripe_customer_id)"
+            " remote(StripeCustomer.stripe_id)=="
+            "foreign(StripeSubscription.stripe_customer_id)"
         ),
     )
     subscription_items = relationship(
--- a/tests/unit/test_crud.py
+++ b/tests/unit/test_crud.py
@@ -1043,7 +1043,6 @@ def test_relations_on_stripe_invoice_line_items(
     assert line_item.get_email_id() == email_id
 
 
-@pytest.mark.xfail
 @pytest.mark.parametrize("contact_idx", range(2))
 @pytest.mark.parametrize("sub_idx", range(2))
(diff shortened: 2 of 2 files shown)


bsieber-mozilla (approved):
LGTM

_The remote/foreign relationships are not super well-documented in the sqlalchemy docs; this seems like a lot of trial and error._

Thanks for tackling this.

PR #336, opened 2021-12-06
If 'other' reset segment-prefix to blank

bsieber-mozilla:
Whenever we receive a `segment` that results in "other" reset the `segment_prefix` to ""


jwhitlock (approved):
Thanks @bsieber-mozilla!

Issue #340, opened 2021-12-08
Database exceptions (IntegrityError, Deadlocks) when updating Stripe objects


jwhitlock:
There are exceptions raised when running ``db.commit()`` as part of processing a payload in ``/stripe_from_pubsub``. Some are due to adding an existing object:

```
IntegrityError('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "stripe_subscription_pkey"
DETAIL:  Key (stripe_id)=(sub_JRf2wXHxlzW2De) already exists.
')
```

Others are deadlocks:

```
OperationalError('(psycopg2.errors.DeadlockDetected) deadlock detected
DETAIL:  Process 27960 waits for ShareLock on transaction 11356748; blocked by process 28164.
Process 28164 waits for ShareLock on transaction 11356749; blocked by process 27960.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (3,14) in relation "stripe_price"
')
```

This implies that the same object is being created or updated in two parallel calls to ``/stripe_from_pubsub``. It may be necessary to use ``SELECT ... FOR UPDATE`` to lock the row when reading if the object already exists. 

Something like this in ``_get_stripe`` may work:

```python
    query = db_session.query(model)
    if for_update:
        query = query.populate_existing().with_for_update()
    return cast(Optional[StripeModel], query.get(stripe_id))
```

The docs for [with_for_update](https://docs.sqlalchemy.org/en/14/orm/query.html?highlight=with_for_update#sqlalchemy.orm.Query.with_for_update) suggest pairing it with ``populate_existing`` to refresh the identity map.

Greedy locking may be needed for subscription items of a subscription, and invoice line items of an invoice.

Additionally, the errors are not being sent to Sentry.  They are viewable in Papertrail by looking for `uvicorn.error`. This may be new with the updates included in v1.1.0

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-119)

jwhitlock:
The Sentry issue was due to a server-side "webcrawler" filter. For whatever reason, the GCP request to ``/stripe_from_pubsub`` is seen as a webcrawler. Turning off this filter allows the exceptions to be registered normally.

jwhitlock:
With v1.1.2 shipped, these are now caught and cause a transaction rollback, with the exception logged (no longer sent to Sentry), and a 409 returned to the caller.

[...]

PR #342, opened 2021-12-09
Update stripe objects more carefully

jwhitlock:
Changes for issue #340:

* Use ``SELECT ... FOR UPDATE``, to try to ensure one writer at a time updates a Stripe row. This is done in the generic ``_get_stripe()`` function, as well as when getting the IDs of subscription items and invoice line items, to detect deletions.
* Use ``filter().one_or_none()`` instead of ``.get()``, to avoid the identity map cache
* Catch ``IntegrityError`` and ``OperationalError`` (deadlocks and other issues), log the exception, and return a 409, instead of a 500.


bsieber-mozilla (approved):
LGTM

Thanks @jwhitlock !

Issue #343, opened 2021-12-09
Duplicate FxA ID leads to repeated PubSub import failures


jwhitlock:
In CTMS stage, a PubSub request leads to an error:

```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint \"ix_stripe_customer_fxa_id\"
DETAIL:  Key (fxa_id)=(0017c480308e47cb95b0488d727019b5) already exists.

[SQL: INSERT INTO stripe_customer (stripe_id, fxa_id, default_source_id, invoice_settings_default_payment_method_id, stripe_created, deleted, update_timestamp) VALUES (%(stripe_id)s, %(fxa_id)s, %(default_source_id)s, %(invoice_settings_default_payment_method_id)s, %(stripe_created)s, %(deleted)s, now())]
[parameters: {'stripe_id': 'cus_GW2467hdiTdGDm', 'fxa_id': '0017c480308e47cb95b0488d727019b5', 'default_source_id': 'card_1Fz00iKb9q6OnNsLCWykJt1i', 'invoice_settings_default_payment_method_id': None, 'stripe_created': datetime.datetime(2020, 1, 9, 12, 21, 13, tzinfo=datetime.timezone.utc), 'deleted': False}]
(Background on this error at: http://sqlalche.me/e/13/gkpj)
```

There is no CTMS contact with this FxA ID, and there are two active two customers in Stripe with the same FxA ID / description.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-118)

jwhitlock:
The FxA Stripe cache keys the customer records by FxA ID, so I believe this customer will either:

1) Be overwritten once, or 
2) Be overwritten with each subscription / invoice update

If 1), a fix is to manually delete the old record (customer ``cus_GW242Wf1U8slA5``) or overwrite it with a new FxA ID (``not_0017c480308e47cb95b0488d727019b5``), probably by ``POST /stripe`` with appropriate data.

Alternatively, we can remove the unique constraint on FxA IDs, especially if we expect this in prod.

jwhitlock:
Since FxA indexes Stripe customers by FxA ID, the duplicate FxA ID will overwrite the existing one in the Fxa Firestore. I think we should do the same. To do this, when updating a stripe customer, we should find existing ``StripeCustomer``s by Stripe ID and by FxA ID:

[...]

PR #345, opened 2021-12-13
Log Stripe ingest actions, and delete existing customer on FxA ID conflict

jwhitlock:
This should fix issue #343, where PubSub is sending a new Stripe customer with the same FxA ID as an existing customer. This is seen on stage, and appears to be due to a bug or direct interaction with the Stripe console. We currently return a 409 Conflict to the FxA PubSub push request, which means it is tried again a few milliseconds later.  Because the FxA Firestore indexes customers by FxA ID, it handles conflicts by overwriting the existing customer. This PR changes the behavior to delete the existing customer (mirroring the overwrite in FxA Firestore) and then import the new customer.

The first part detects the conflicting FxA ID in the application (``ingest_stripe_customer``) rather than during the database commit. For new customers, an extra ``SELECT`` is executed to find matching rows by FxA ID. For updated customers, the ``SELECT`` is only executed if the FxA ID changed. When this happens, ``ingest_stripe_customer`` raises a new exception, ``StripeIngestFxAIDConflict``.

The second part is to handle this exception. This is done in the shared code for the ``/stripe`` and ``/stripe_from_pubsub`` handlers. The existing customer is deleted, and ``ingest_stripe_customer`` is called again, to ingest the new stripe customer without conflict. Stripe data related to the deleted customer, such as subscriptions and invoices, is **not** deleted, due to our data model of weak relations between Stripe objects.

To report on this new action, there are two new items in the log context for Stripe API calls (``/stripe`` and ``/stripe_from_pubsub``), ``fxa_id_conflict`` and ``ingest_actions``.

The first is ``fxa_id_conflict``, which is set when a new Stripe customer has the same FxA ID (in the ``description`` field) as an existing customer.  If there is no conflict, this item is omitted.

The second is ``ingest_actions``, a dictionary of actions to a list of objects. For example, a subscription update may look like: 

```json
{
  "created": [
    "price:plan_XZY",
    "subscription:sub_abc123",
    "subscription_item:si_abc123_xzy"
  ]
}
```
The possible actions are:

* ``created``: A new object was ingested
* ``updated``: An existing object was updated
* ``no_change``: An existing object was ingested, but no changes in the data CTMS cares about
* ``deleted``: An existing object was deleted
* ``skipped``: The ingested object was skipped and not added to the database. One example is a new customer that is marked as deleted in Stripe - CTMS does not bother creating the customer row.

The log context ``stripe_unknown_objects``, which listed unknown Stripe object types, is removed. Instead, these are logged under ``ingest_actions["skipped"]``.

``ingest_actions`` gives a way to identify the conflicting customer records. The existing one will be listed under ``ingest_actions["deleted"]``, and the new one under ``ingest_actions["created"]``.  This also gives some insight into the PubSub push messages, so we can see what kind of data is in a request, and what is in a sequence of requests. Getting the ingest actions up to the application required changing the return values of ingest functions like ``ingest_stripe_customer``, so there are many repetitive changes. If it wasn't so late in the year, I'd refactor this feature into its own PR.

The ``ingest_actions["no_change"]`` items are implemented by looking at the ORM objects before and after the update, and determining if any of the fields changed. This exposed some bugs in updating that are fixed:

* The unfortunately named [datetime.utcfromtimestamp()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp) returns a timezone naive ``datetime``, causing all changes to look like updates. This switches to ``datetime.fromtimestamp(timestamp, tz=timezone.utc)``. I think we got burned in a similar way with ``datetime.utcnow()``. Thanks to the blog post [Stop using utcnow and utcfromtimestamp](https://blog.ganssle.io/articles/2019/11/utcnow.html).
* Fix update of ``invoice.default_source_id``.
* Fix update of ``invoice_line_item.stripe_subscription_item_id``.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -46,7 +46,11 @@ from .crud import (
 )
 from .database import get_db_engine
 from .exception_capture import init_sentry
-from .ingest_stripe import StripeIngestUnknownObjectError, ingest_stripe_object
+from .ingest_stripe import (
+    StripeIngestActions,
+    StripeIngestUnknownObjectError,
+    ingest_stripe_object,
+)
 from .log import configure_logging, context_from_request, get_log_line
 from .metrics import (
     emit_response_metrics,
@@ -903,7 +907,7 @@ def metrics(request: Request):
 
 def _process_stripe_object(
     db_session: Session, data: Dict
-) -> Tuple[Optional[UUID], Optional[str]]:
+) -> Tuple[Optional[UUID], Optional[str], StripeIngestActions]:
     """
     Ingest a Stripe Object and extract related data.
 
@@ -918,7 +922,7 @@ def _process_stripe_object(
       data for keys that CTMS examines. Extra data is ignored.
     """
     try:
-        obj = ingest_stripe_object(db_session, data)
+        obj, actions = ingest_stripe_object(db_session, data)
         db_session.commit()
     except IntegrityError as e:
         db_session.rollback()
@@ -936,7 +940,7 @@ def _process_stripe_object(
         trace_email = data["email"]
     else:
         trace_email = None
-    return email_id, trace_email
+    return email_id, trace_email, actions
 
 
 @app.post(
@@ -953,7 +957,7 @@ def stripe(
     if not ("object" in data and "id" in data):
         raise HTTPException(status_code=400, detail="Request JSON is not recognized.")
     try:
-        email_id, trace_email = _process_stripe_object(db_session, data)
+        email_id, trace_email, actions = _process_stripe_object(db_session, data)
     except (KeyError, ValueError, TypeError) as exception:
         raise HTTPException(
             400, detail="Unable to process Stripe object."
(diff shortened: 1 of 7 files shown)


bsieber-mozilla (approved):
> Stripe data related to the deleted customer, such as subscriptions and invoices, is not **deleted**, due to our data model of weak relations between Stripe objects.

Is there any concern for it not to be deleted? (I'm confused whether these leftover objects are to be reused or just remain.)


The `ingest_actions` context looks awesome, thanks for that detail!

Those dangerous datetime methods; the fix and reference link are much appreciated!

--
Thanks for this pr @jwhitlock!

jwhitlock:
> Is there any concern for it not to be deleted? (I'm confused whether these leftover objects are to be reused or just remain.)

I'm not sure what happens on the FxA side for these Customers with duplicate FxA IDs. Does one win forever? Or does the same FxA Firestore get overwritten depending on which one has the most recent change? If it gets overwritten, the orphan subscriptions will find their customer again.

It appears to just affect two Stripe customers in stage, so I think the data accumulation will be minimal. There will be other data accumulation, and we may want to periodically find and delete stale data. I've created https://github.com/mozilla-it/ctms-api/issues/346 to track that.

PR #347, opened 2021-12-15
Update lxml from 4.6.4 to 4.7.1

jwhitlock:
There are a few changes:
 
https://lxml.de/4.7/changes-4.7.1.html

lxml 4.6.5 includes a security fix in the HTML cleaner. I do not believe that CTMS uses this feature, so we're not vulnerable. We use `lxml.etree` to parse the Acoustic response, and none of our dependencies use `lxml`.


bsieber-mozilla (approved):
LGTM thanks @jwhitlock

PR #349, opened 2021-12-16
Updating lines from cinchy/basket 

bsieber-mozilla:
Updating lines from cinchy/basket to avoid incorrect presentation of communication between aformentioned systems.

_Previously lines from cinchy/basket connected, this PR addresses representation concerns that this could introduce._


jwhitlock (approved):
Thanks @bsieber-mozilla, good change!

PR #351, opened 2021-12-17
Handle POST of new deleted customer

### LINKED ISSUE #350: Stripe ingest error when a new customer is deleted
When a Stripe customer is deleted, the payload is:

```json
{
  "deleted": true, 
  "id": "cus_KiMjcNh2r0LRic",
  "object": "customer"
}
```

If CTMS does not have a record for this customer already, it skips ingesting, doesn't create a record, and returns `obj -= one`. However, the API code then tries to call ``obj.get_email_id()`` and fails. This returns a 500 Server Error to PubSub, which then tries again with the same result...

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-125)

jwhitlock:
Handle the case where a new Stripe customer starts out deleted. The ingest is successful (doesn't raise `StripeIngestUnknownObjectError`), but no object is created and the object returned is `None`.

Fixes #350. Since this is a production issue, also preparing for next release.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -961,7 +961,7 @@ def _process_stripe_object(
             status_code=409, detail="Deadlock or other issue, try again"
         ) from e
 
-    email_id = obj.get_email_id()
+    email_id = obj.get_email_id() if obj else None
     if data["object"] == "customer" and re_trace_email.match(data.get("email", "")):
         trace_email = data["email"]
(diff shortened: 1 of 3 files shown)


bsieber-mozilla (approved):
Thanks @jwhitlock

PR #352, opened 2022-01-18
Add RelayWait list table

leplatrem:


bsieber-mozilla:
This and the other PR both seem to be following the same method of implementation; are you ok with moving forward with the other one?

leplatrem:
Of course!

PR #353, opened 2022-01-19
Adding Relay Waitlist; following VPN Waitlist pattern at this time.

bsieber-mozilla:
@leplatrem I believe this handles the majority of required changes to construct another waitlist (similar to the `VPNWaitlist`). 

Let me know what you think


leplatrem:
Ha! I had a similar result in #352 so I guess this looks good to me :)

bsieber-mozilla:
Ok--if we're satisfied--I believe these are the next steps:
- [ ] Add new field(`relay_waitlist_geo`) into main database table in Acoustic
  - [ ] Use the `+add field`-button on the database 
- [ ] Confirm basket requests have new additions to schema
  - [ ] ex: `{'relay_waitlist':{'geo': 'xy'}}`
  
_Pending approval._

jwhitlock (approved):
Thanks, looks good to me!

leplatrem (approved):
I obviously can't tell whether this patch touches all relevant parts exhaustively, but the changes surely look sound, so r+ :man_shrugging:

jwhitlock:
I did think of an issue with this PR. I'm not entirely sure what the deploy sequence is, with respect to migrations. It could be this:

1. Run migrations
2. Start up pods with the new code
3. As new code pods report as ready, add to load balancer and remove one old pod
4. When all traffic is going to new pods, terminate the old ones

We'd be OK in this state - the new database tables would be applied before the new code runs. The old code would continue to work because it would ignore the new table.

If it looks like this though, we're in trouble:

1. Start up pods with the new code
2. Run migrations
3. As new code pods report as ready, add to load balancer and remove one old pod
4. When all traffic is going to new pods, terminate the old ones

... because the new pods will try to access a database that isn't there.

No matter the deploy strategy, breaking up the PR will make it work:

1. Deploy the alembic migration that updates the database, but not the code or model change. The table is now present but unused.
2. Deploy the code that uses the new table

bsieber-mozilla:
Converted this PR to a draft. I split the efforts in the two referenced draft PRs: #358 , #359

Issue #354, opened 2022-01-23
PubSub ingester chokes on null value in payload


jwhitlock:
This line of the pubsub ingester:

https://github.com/mozilla-it/ctms-api/blob/904a8e40fc3e312dabc769d0742e6668d18f4be4/ctms/app.py#L1030

is raising an exception:

```python
TypeError("argument of type 'NoneType' is not iterable")
```

It is expecting a payload dictionary like:

```json
{
   "firestore_table:stripe_id": {
      "object": "stripe_type",
      "other": "values"
   }
}
```

This suggests a payload dictionary like:

```json
{
   "obj:obj_id": null
}
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-126)

jwhitlock:
On the CTMS side, something like this may work:

```python
    if hasattr(payload, "keys"):
        if all(key.count(":") == 1 for key in payload.keys()):
            # Item dictionary, "firebase_table:id" -> Stripe object
            items = []
            for key, value in payload.items():
                if value:
                    items.append(value)
                else:
                    logger.warning("PubSub key %s had empty value %s", key, value)
        else:
            items = [payload]  # One Stripe object, or maybe unknown dictionary
```

Alternatively, the pubsub cloud function could discard null values.

We may need to coordinate with the FxA team to determine why null values are being added to Firestore.

bsieber-mozilla:
Thanks for looking into and outlining this issue; I've implemented your proposed solution in PR #355 

I think keeping an eye on the amount of `None` we're receiving could be insightful. It looks like there was only one triggered incident; one message being held up.

bsieber-mozilla:
I believe we can close this now.

PR #355, opened 2022-01-24
Resolve when receiving None value

bsieber-mozilla:
This PR is looking to address issue #354 and uses the proposed solution by @jwhitlock

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -1,5 +1,6 @@
 """The CTMS application, including middleware and routes."""
 # pylint:disable = too-many-lines
+# pylint:disable = too-many-statements
 import json
 import sys
 import time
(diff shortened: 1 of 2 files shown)


bsieber-mozilla:
Next steps:  

- [ ] Bump version ( #357 )
- [ ] Deploy new version

ctms/app.py
```diff
@@ -1,5 +1,6 @@
 """The CTMS application, including middleware and routes."""
 # pylint:disable = too-many-lines
+# pylint:disable = too-many-statements
```
jwhitlock, line 3: ouch! :hurtrealbad:

It may be worth breaking this into more files, but definitely outside of this PR.

jwhitlock (approved):
Thank you @bsieber-mozilla !

ctms/app.py
bsieber-mozilla: https://github.com/mozilla-it/ctms-api/issues/360

Constructed that issue to keep track of this, thanks for the call out.

PR #356, opened 2022-01-24
Update codeowners to be SysEng

bsieber-mozilla:
@jwhitlock what are you thoughts on this?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 # Thanks to @imbstack for the suggestion.
 # see http://code.v.igoro.us/posts/2019/07/codeowners.html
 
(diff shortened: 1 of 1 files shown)


jwhitlock:
I think it would be a good agenda items for the next SysEng team meeting. I think https://github.com/orgs/mozilla-services/teams/syseng is a good list for access, and could also be the review list. I think it would be worth confirming if everyone is:

* Willing to be in the review rotation
* Has SAML login setup for mozilla-it

I'm not sure if a certain level of access is needed for your review to "count" for unblocking a merge. This PR might be a good test if existing people's review counts.

bsieber-mozilla:
Thinking a bit more on this; we might want a new team dedicated to CTMS (reviewers)?

PR #368, opened 2022-02-14
Adding relay waitlist to models and apis...

bsieber-mozilla:
Minor adjustments to previous PR #359

### DIFF
--- a/tests/unit/test_acoustic_service.py
+++ b/tests/unit/test_acoustic_service.py
@@ -59,8 +59,8 @@ def test_ctms_to_acoustic(
     base_ctms_acoustic_service, example_contact, maximal_contact, minimal_contact
 ):
     contact_list = [example_contact, maximal_contact, minimal_contact]
-    example_contact_expected = [50, len(example_contact.newsletters) - 2]
-    maximal_contact_expected = [50, len(maximal_contact.newsletters) - 2]
+    example_contact_expected = [51, len(example_contact.newsletters) - 2]
+    maximal_contact_expected = [51, len(maximal_contact.newsletters) - 2]
     minimal_contact_expected = [32, len(minimal_contact.newsletters) - 2]
     expected_results = {
         example_contact.email.email_id: example_contact_expected,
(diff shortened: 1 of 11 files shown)


tests/unit/test_acoustic_service.py
```diff
@@ -59,8 +59,8 @@ def test_ctms_to_acoustic(
     base_ctms_acoustic_service, example_contact, maximal_contact, minimal_contact
 ):
     contact_list = [example_contact, maximal_contact, minimal_contact]
-    example_contact_expected = [50, len(example_contact.newsletters) - 2]
-    maximal_contact_expected = [50, len(maximal_contact.newsletters) - 2]
```
leplatrem, line 63: Can we test somewhere that the relay waitlist info will land in Acoustic?

leplatrem (approved):
The changes of this PR look sound with regards to what was done for the VPN.

But seeing the number of required changes to onboard this use-case, and the overall level of abstraction of this code, it is pretty clear that we can't follow this approach for many other waitlists!

tests/unit/test_acoustic_service.py
bsieber-mozilla: By this do you mean testing the flow from CTMS(Stage) to Acoustic (Sandbox)?... or prior to merging/local tunneling?

I imagine, you could attempt to simulate a [sync](https://github.com/mozilla-it/ctms-api/blob/cde9694e2992c2c9d8315ba9c6954c50a6facf37/ctms/bin/acoustic_sync.py) if you've got the associated [env vars setup](https://github.com/mozilla-it/ctms-api/blob/cde9694e2992c2c9d8315ba9c6954c50a6facf37/ctms/config.py#L71-L79).

I've yet to do such a task, perhaps there's a better route?

The field exists in Acoustic Sandbox now; but we will need to ensure it's added by Marketing for Production.
Screen Shot 2022-03-17 at 4 17 39 PM

leplatrem: Testing with a real sandbox would be ideal, but at least mock the client and/or inspect the calls to Acoustic in unit tests

bsieber-mozilla: Created issue #370 to reference this

PR #371, opened 2022-03-28
Architecture change proposal: Basket inbox endpoint

leplatrem:


bsieber-mozilla:
I really like this suggestion, thanks for proposing it

I'm confused as to how Option 1 would be "Fuzzy" in terms of separation of concern; from my point of view--it seems that by providing the full data blob to CTMS, Basket no longer needs to be concerned with: the format, as well as what "newsletters" are actually not newsletters. 

I think this allows Basket to simplify the pre-existing logic (which was previously there for SFDC) and migrate it to the corresponding system. It feels odd that Basket receives the data, converts the flat structure to a hierarchical one, and then in CTMS we end up flattening the structure to meet the needs of Acoustic--this proposition could allow us to look more closer at our interactions and expectations for using Acoustic.

leplatrem:
> I'm confused as to how Option 1 would be "Fuzzy" in terms of separation of concern; 

I wanted to express that CTMS now has to be aware of what's coming from Basket. Whereas before, Basket was in charge of complying with the CTMS API. In an ideal world, each system complies with the remote service, not the other way around.  

From the CTMS point of view, we thus add a new «concern»: the schema of the incoming data from Basket.

The only reason why we are considering breaking this separation of concern is to reduce the number of systems to be changed when onboarding a new usecase.

Maybe «fuzzy» is not the right word here

bsieber-mozilla:
> I wanted to express that CTMS now has to be aware of what's coming from Basket. Whereas before, Basket was in charge of complying with the CTMS API. In an ideal world, each system complies with the remote service, not the other way around.

That makes sense. I think initially CTMS will be more aware of the structure and might have some "fuzzy" overlap of internal knowledge of the other system, but I imagine/hope that the processing of these payloads to change in the future--or at least the process for waitlists. 

I think this proposition gives power to CTMS to make a change to this paradigm for how we're collecting and transforming the data.

leplatrem:
@mozilla-it/ctms-reviewers should we drop the pull-request, or does it make sense to merge it with `Status: abandonned` to keep track of past and discarded ideas?

PR #373, opened 2022-04-07
Updates to Pre-commit and DB changes 

bsieber-mozilla:
Looking to resolve: 
https://mozilla-hub.atlassian.net/jira/software/c/projects/CTMS/boards/364
https://sentry.prod.mozaws.net/operations/ctms-stage/issues/15860536/?query=is%3Aunresolved
https://sentry.prod.mozaws.net/operations/ctms-stage/issues/19094618/?query=is%3Aunresolved

Slack thread with some additional details:
https://mozilla.slack.com/archives/G01M9NUSUG1/p1648750262524159

### DIFF
--- a/ctms/config.py
+++ b/ctms/config.py
@@ -19,6 +19,9 @@ class LogLevel(str, Enum):
 
 class Settings(BaseSettings):
     db_url: PostgresDsn
+    db_pool_size: int = 10
+    db_max_overflow: int = 20
+    db_pool_timeout_in_seconds: int = 45
     secret_key: str
     token_expiration: timedelta = timedelta(minutes=60)
     server_prefix: str = "http://localhost:8000"
--- a/ctms/database.py
+++ b/ctms/database.py
@@ -6,7 +6,12 @@ from . import config
 
 
 def get_db_engine(settings: config.Settings):
-    engine = create_engine(settings.db_url)
+    engine = create_engine(  # https://docs.sqlalchemy.org/en/14/core/engines.html#engine-creation-api
+        settings.db_url,
+        pool_size=settings.db_pool_size,
+        max_overflow=settings.db_max_overflow,
+        pool_timeout=settings.db_pool_timeout_in_seconds,
+    )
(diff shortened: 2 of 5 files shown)


ctms/config.py
```diff
@@ -19,6 +19,9 @@ class LogLevel(str, Enum):
 
 class Settings(BaseSettings):
     db_url: PostgresDsn
+    db_pool_size: int = 10
+    db_max_overflow: int = 20
+    db_pool_timeout_in_seconds: int = 45
```
leplatrem, line 24: These do not match the defaults of SQLA, how did you pick these values?

leplatrem (approved):
These new settings may deserve a mention in `docs/configuration.md`?

ctms/config.py
bsieber-mozilla: After viewing:
https://my.papertrailapp.com/groups/24734221/events?q=exception&focus=1448826046895013939&selected=1448826046895013939
```
TimeoutError('QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30')
```
and looking into: https://docs.sqlalchemy.org/en/13/errors.html#queuepool-limit-of-size-x-overflow-y-reached-connection-timed-out-timeout-z

The defaults were pool_size=5, max_overflow=10, pool_timeout=30; since these values were not working and were resulting in both QueuePool limit's and timeout's being hit--upping these values made sense to me. The pool_size and max_overflow were doubled and the timeout was increased by 50%. 

Do you suggest a particular approach for picking these values?

leplatrem: I'm ok, hence the approval ;) 

I didn't have the backstory, and just wanted to underline that there were 2 intentions here: making it configurable, and increasing defaults

bsieber-mozilla: That makes sense, those are the two things occurring; thanks for calling it out!

bsieber-mozilla:
Thanks for the review @leplatrem! I've added these new env var's to the configuration document and will merge this in

PR #375, opened 2022-04-12
When 'for_update' is True, use the 'model' provided as the 'of' param to 'query.with_for_update'

bsieber-mozilla:


bsieber-mozilla:
Moving this forward since @leplatrem  and I chatted about it earlier. 

_Can revert if need be._

PR #377, opened 2022-04-13
Removing 'of' param in call to 'with_for_update'

bsieber-mozilla:
Modifying parameters used during the locking of rows.

https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.with_for_update

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -724,7 +724,7 @@ def _get_stripe(
     """
     query = db_session.query(model)
     if for_update:
-        query = query.with_for_update(of=StripeBase)
+        query = query.with_for_update()
     return cast(
         Optional[StripeModel], query.filter(model.stripe_id == stripe_id).one_or_none()
     )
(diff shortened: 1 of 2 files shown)


bsieber-mozilla:
While looking into modifying the locking param-model; I've delved into the Push Subscription settings to modify:
- increased ACK deadling to 30s (matching timeout from sqlalchemy DB calls)
- modified retry policy--from immediately to exponential backoff.

If a customer document is updated--and underlying collections are updated--since there are trigger functions on multiple levels of the Firestore collections hierarchies, the single stripe event/update might result in multiple requests with the same data to CTMS. 

It could be possible we've got many messages all trying to update and lock the same DB rows--causing a multitude of problems, including Deadlock's.

bsieber-mozilla:
_This reverts changes from #375_

leplatrem (approved):
Approving the revert

bsieber-mozilla:
Thanks for the Approval; going to hold off on this merge for the time being... since the queue's drain-rate is faster than it's fill-rate after infra changes made in the PS instance.

PR #380, opened 2022-04-18
Adding environment variable for sql_alchemy create_engine call pool_recycle: `db_pool_recycle_in_seconds`

bsieber-mozilla:
https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine.params.pool_recycle

Adding this param to create_engine; with default of 5min.


bsieber-mozilla:
Hypothesis> this should "kick" the connections after 5minutes of recycled--usage.

PR #383, opened 2022-04-19
Upping time to 15 min

bsieber-mozilla:


leplatrem:
wouldn't it be easier to change the configuration rather than the application defaults?

bsieber-mozilla:
In the long run, yes; initially we should determine what a good default could be, I think modifying this config is the initial approach

PR #384, opened 2022-04-19
chore(stripe): Set for_update to False on price query

bkochendorfer:
This data should not change very often and we are locking multiple times causing the entire application to lock up.


bsieber-mozilla (approved):
Thanks for this PR; lgtm @bkochendorfer

_let's see how it goes_

PR #386, opened 2022-05-19
Add script to parse stripe exports from firestore and add them to acoustic queue.

bsieber-mozilla:


bsieber-mozilla:
This script is intended to use the same (munged) firestore export that ctms/bin/ingest_stripe_data.py works with; and as such follows a similar pattern. 

This is rolled into it's own script instead of into the ingest_stripe_data.py script to separate and make the processes explicit.

bkochendorfer (approved):

PR #392, opened 2022-06-06
Adding logging for stripe ingestion

bsieber-mozilla:


grahamalama (approved):
Lgtm. There might be room to play with the formatting of the logging string to make it more readable? But whatever the format, having that additional context is useful.

PR #411, opened 2022-08-15
Only push Docker images on commits to `main` branch

grahamalama:
This PR modifies our Actions so that we only upload built containers to the container registry on pushes to `main` (usually during PR merges or pushed tags).

This is (among other things) so that Dependabot PRs can pass CI. They're currently failing because they don't have access to our AWS creds, but we don't _need_ to login and push the containers when we run the test suite.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -1,7 +1,12 @@
 name: Build Docker image
 
-on: [ push, pull_request ]
-
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+      - stg-[0-9a-f]{6,10} 
 jobs:
   prepare:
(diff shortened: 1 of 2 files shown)


bkochendorfer (approved):
Cool good work @grahamalama !

grahamalama:
@bkochendorfer looks like this PR is stuck on waiting for `prepare` and `build-image`. Could you maybe force merge this PR? Hopefully this won't happen after this gets merged.

bkochendorfer:
hmm, I don't have special powers here to force merge this

bsieber-mozilla:
It looks like I can "Merge as Admin", is that what we're looking for here. 

All aboard? 🚢

bkochendorfer:
Sounds good @bsieber-mozilla if there is an issue with other PRs we can figure out what to do using your magic admin powers

PR #425, opened 2022-10-05
[CTMS-136] Adds a new newsletter for Graceland

grahamalama:
In [CTMS-136](https://mozilla-hub.atlassian.net/browse/CTMS-136), a new waitlist was requested for the `Graceland` project. After talking with Stephen Hood, we determined that a plain newsletter was sufficient to address his usecase, rather than a `waitlist` a la #368.

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -106,6 +106,7 @@ class AcousticResources:
         "firefox-accounts-journey": "sub_firefox_accounts_journey",
         "firefox-news": "sub_firefox_news",
         "firefox-sweepstakes": "sub_firefox_sweepstakes",
+        "graceland": "sub_graceland",
         "hubs": "sub_hubs",
         "internet-health-report": "sub_internet_health_report",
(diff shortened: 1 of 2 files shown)


ctms/acoustic_service.py
```diff
@@ -106,6 +106,7 @@ class AcousticResources:
         "firefox-accounts-journey": "sub_firefox_accounts_journey",
         "firefox-news": "sub_firefox_news",
         "firefox-sweepstakes": "sub_firefox_sweepstakes",
+        "graceland": "sub_graceland",
```
bsieber-mozilla, line 109: Has this field "sub_graceland" been made in Acoustic?

bsieber-mozilla (approved):
LGTM; thanks!

ctms/acoustic_service.py
bsieber-mozilla: _(We can make this in staging and production with the "+ add field" option, I believe.)_

grahamalama:
Going to close this since we've retired this approach with #456

PR #426, opened 2022-10-07
Temporarily disable / comment out "For Update" in SQL queries

grahamalama:
This is only to clear the pubsub queue we're seeing on staging.  This should be reverted when the issue is resolved.


bsieber-mozilla (approved):
LGTM

PR #432, opened 2022-10-12
Revert `FOR UPDATE` commits

grahamalama:
Reverts commits that removed `FOR UPDATE` in various Stripe-related queries. We still want to investigate these queries to determine if we can remove these clauses, but the commenting out / flipping arguments was a temporary measure to fix an issue in staging.

Reverts:
- d75bbe4f31e07a03178179e93137494031ce830d
- 5f7d40a04b6cb0b8d83bb2983ff1de125a54f133
- b008370a2ee9f43837d800a60b758d96deaf80dc

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -723,8 +723,8 @@ def _get_stripe(
     If for_update is True (default False), the row will be locked for update.
     """
     query = db_session.query(model)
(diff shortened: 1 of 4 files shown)


bsieber-mozilla (approved):
🚢

PR #435, opened 2022-10-17
Refactor deletion of orphaned line items and subscription items

grahamalama:
This PR:

- (hopefully) optimizes the `ingest_stripe_invoice` function. Instead of using `SELECT ... FOR UPDATE` to fetch invoice line items before we needed them, then using those line items later in the function to determine which line items (if any) to delete, we add another clause to the `DELETE` query to achieve the same outcome.
- Removes assertions about the contents of SQL queries in tests. These assertions felt like implementation details in the context of the `ingest_invoice_data` function. We should assert that the `for_update` kwarg does what we want and (maybe) that we don't encounter any integrity errors when trying to do updates, but asserting certain SQL fragments are in the statements feels too brittle.

### DIFF
--- a/ctms/ingest_stripe.py
+++ b/ctms/ingest_stripe.py
@@ -19,6 +19,8 @@ from __future__ import annotations
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Set, Tuple, cast
 
+from sqlalchemy import delete
+
 from ctms.crud import (
     StripeModel,
     create_stripe_customer,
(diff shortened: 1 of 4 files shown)


ctms/ingest_stripe.py
```diff
@@ -350,15 +371,6 @@ def ingest_stripe_invoice(
         invoice.status = data["status"]
         invoice.default_payment_method_id = data["default_payment_method"]
         invoice.default_source_id = data["default_source"]
-        lines_to_delete = {
-            line_id
-            for line_id, in (
-                db_session.query(StripeInvoiceLineItem.stripe_id)
-                .with_for_update()
-                .filter(StripeInvoiceLineItem.stripe_invoice_id == invoice.stripe_id)
-                .all()
```
grahamalama, line 359: Not only did this query fetch line items before we used them, but it also fetched more than we needed, since it fetches all of the line items, then later we remove line items from the set that appear in the incoming invoice data.

bsieber-mozilla (approved):
🚢

ctms/ingest_stripe.py
```diff
@@ -327,6 +329,25 @@ def ingest_stripe_price(
...
+    """Remove any orphaned invoice items, returning IDs of the deleted items
+
+    Orphaned line items are line items that exist on the invoice that don't
+    exist in the incoming line item data.
+    """
+    stmt = (
+        delete(StripeInvoiceLineItem)
+        .where(StripeInvoiceLineItem.stripe_invoice_id == invoice.stripe_id)
+        .where(StripeInvoiceLineItem.stripe_id.notin_(line_item_data_ids))
+        .execution_options(synchronize_session=False)
```
leplatrem, line 344: Can you add a comment specifying that we're not interested in synchronizing objects here?

leplatrem (approved):
LGTM

bkochendorfer:
:nail_care: Nice

leplatrem (approved):

grahamalama:
Elements of this PR might be useful in the future, but the biggest performance improvement was gained by adding an index to the right table. #445

PR #442, opened 2022-10-19
Aligning version practices throughout services. Removing python packa…

bsieber-mozilla:
…ge versioning, setting internal defaults to 0/v0.0.0, which are overwritten by git-tags.

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -86,10 +86,12 @@ from .schemas import (
     VpnWaitlistSchema,
 )
 
+version_info = get_version()
+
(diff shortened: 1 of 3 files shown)


bsieber-mozilla:
_- Next up, have root show versioning info; as opposed to rerouting to /docs [(similar pattern as JBI)](https://github.com/mozilla/jira-bugzilla-integration/blob/d415fc023b115df30482c1606a9c2e7b283b3fbe/jbi/router.py#L21)_

leplatrem (approved):

PR #448, opened 2022-11-11
Store list of Acoustic fields in the database

leplatrem:
- [x] Add a new table
- [x] Populate it with current hardcoded values
- [x] Replace hardcoded values with table read
- [x] Update tests
- [x] Add command / API ? to CRUD values


docs/deployment_guide.md
```diff
@@ -55,6 +55,19 @@ from this repo and going outwards:
 More information about CTMS operations is available on the
 [ESS-CTMS Mana page](https://mana.mozilla.org/wiki/x/KIyXC) (*private*).
 
+
+## Synchronized Acoustic Fields
```
bsieber-mozilla, line 59: Thanks for adding this documentation!

migrations/versions/20221111_06f24cc01c09_add_acoustic_field.py
```diff
@@ -0,0 +1,84 @@
...
+# revision identifiers, used by Alembic.
+revision = '06f24cc01c09'  # pragma: allowlist secret
+down_revision = '406251207612'  # pragma: allowlist secret
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    acoustic_field_table = op.create_table('acoustic_field',
```
bkochendorfer, line 26: Since this has moved from being in memory to a database call and we seem to be querying it frequently (I think correct me here if I am wrong). Should we be concerned with database performance or adding any indexes here?

tests/unit/test_sync.py
```diff
@@ -135,12 +136,14 @@ def test_sync_acoustic_record_delete_path(
...
-        assert watcher.count == 4  # Get All Records, Get Contact(x2), Increment Retry
+        assert (
+            watcher.count == 5
+        )  # Get Acoustic Fields, All Records, Get Contact(x2), Increment Retry
         assert context == expected_context
         return
 
     # Metrics adds two DB queries (total records and retries)
-    assert watcher.count == 6
+    assert watcher.count == 7
```
grahamalama, line 146: Is it worth having assertions like this, or is this something we can safely delete?

tests/unit/conftest.py
```diff
@@ -77,8 +83,10 @@ def engine(pytestconfig):
     echo = pytestconfig.getoption("verbose") > 2
     test_engine = create_engine(test_db_url, echo=echo)
 
-    # TODO: Convert to running alembic migrations
-    Base.metadata.create_all(bind=test_engine)
+    cfg = alembic_config.Config(os.path.join(APP_FOLDER, "alembic.ini"))
+    with test_engine.begin() as cnx:
+        cfg.attributes["connection"] = cnx  # pylint:disable = E1137
```
grahamalama, line 88: Can we use the long name here (unsupported-assignment-operation) instead of the error code?

Also, is there a way we could avoid using this `pylint:disable` in the first place?

tests/unit/conftest.py
```diff
@@ -44,6 +46,10 @@
 )
 from tests.unit.sample_data import SAMPLE_CONTACTS, SAMPLE_STRIPE_DATA
 
+MY_FOLDER = os.path.dirname(__file__)
+TEST_FOLDER = os.path.dirname(MY_FOLDER)
+APP_FOLDER = os.path.dirname(TEST_FOLDER)
```
grahamalama, line 51: ✅ nice

ctms/bin/acoustic_fields.py
```diff
@@ -0,0 +1,77 @@
...
+            dbsession.query(AcousticField)
+            .filter(
+                AcousticField.tablename == args.tablename,
+                AcousticField.field == args.field,
+            )
+            .one_or_none()
+        )
+        if not row:
+            print(f"Unknown field '{args.tablename}.{args.field}'. Give up.")
+            return 2
```
grahamalama, line 58: Should we return one of the exit codes as defined in the `os` module? [E.g.](https://docs.python.org/3/library/os.html#os.EX_DATAERR)

ctms/bin/acoustic_fields.py
```diff
@@ -0,0 +1,77 @@
...
+        dbsession.delete(row)
+        dbsession.commit()
+        print("Removed.")
+    else:
+        entries = dbsession.query(AcousticField).all()
+        print("\n".join(sorted(f"- {e.tablename}.{e.field}" for e in entries)))
+
+    return 0
```
grahamalama, line 66: maybe `return os.EX_OK`?

migrations/versions/20221111_06f24cc01c09_add_acoustic_field.py
leplatrem: Thanks for the reminder.
We fetch the whole content of the table once when the `ctms/bin/acoustic_sync.py` script is executed. 
Hence we're good here

tests/unit/test_acoustic_service.py
```diff
@@ -63,18 +63,9 @@ def test_ctms_to_acoustic(
...
-        20 + len(main_acoustic_fields),
-        len(maximal_contact.newsletters) - 2,
-    ]
-    minimal_contact_expected = [
-        18 + len(main_acoustic_fields),
-        len(minimal_contact.newsletters) - 2,
-    ]
+    example_contact_expected = [52, len(example_contact.newsletters) - 2]
+    maximal_contact_expected = [52, len(maximal_contact.newsletters) - 2]
+    minimal_contact_expected = [33, len(minimal_contact.newsletters) - 2]
```
grahamalama, line 68: These assertions always struck me as a bit magic. Any way we could rework this so that things are less flaky and/or more obvious?

migrations/versions/20221111_06f24cc01c09_add_acoustic_field.py
bkochendorfer: Great thank you!

bsieber-mozilla (approved):

tests/unit/test_sync.py
leplatrem: I added an operation that introduces a new SQL statement. Therefore I incremented the value and modified the code comment.

We could invest some time in the unit tests, but I personally would prefer if we wouldn't block this PR for that

tests/unit/conftest.py
leplatrem: > Also, is there a way we could avoid using this pylint:disable in the first place?

Pylint does not understand what is going here:

https://github.com/sqlalchemy/alembic/blob/7a516ebefb8f6bf9041c51aa8711fa2a3a3f4bcd/alembic/config.py#L137-L155

leplatrem: I used the error code to avoid wrapping

```diff
-        cfg.attributes["connection"] = cnx  # pylint:disable = E1137
+        cfg.attributes[  # pylint:disable = unsupported-assignment-operation
+            "connection"
+        ] = cnx
```

tests/unit/test_acoustic_service.py
leplatrem: In the end, the PR does not modify these tests

tests/unit/conftest.py
grahamalama: What about 
```python
        # pylint: disable-next=unsupported-assignment-operation
        cfg.attributes["connection"] = cnx
```

PR #468, opened 2022-11-25
Fix #461: Add `/acoustic_configuration` endpoint to expose Acoustic fields configuration

### LINKED ISSUE #461: Expose mapped newsletters and synced Acoustic fields through API endponits
We recently moved CTMS -> Acoustic newsletter mapping and synced Acoustic fields from constants defined in the source code to tables in the database. (#448 & #456)

In order to inspect the currently mapped newsletters and synced Acoustic fields, we should expose these through the application in some way.

leplatrem:
Fix #461

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -39,9 +39,9 @@ from .config import re_trace_email
 from .crud import (
     create_contact,
     create_or_update_contact,
-    get_api_client_by_id,
     get_all_acoustic_fields,
     get_all_acoustic_newsletters_mapping,
+    get_api_client_by_id,
     get_bulk_contacts,
     get_contact_by_email_id,
     get_email,
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -919,12 +919,12 @@ def metrics(request: Request):
     return Response(generate_latest(registry), status_code=200, headers=headers)
 
 
-@app.get("/configuration", tags=["Platform"])
+@app.get("/acoustic_configuration", tags=["Platform"])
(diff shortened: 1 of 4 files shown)


leplatrem:
> `/configuration` maybe feels a bit generic for this endpoint.

I felt that we were going to expose more stuff there. But yeah you're right, YAGNI

> Also, curious why you chose to combine the mapped newsletters and synced fields in one endpoint instead of an endpoint for each. Not saying this approach needs to change, but they feel like separate concerns.

My idea was to limit the number of endpoints (less noise, less documentation), and have one endpoint that exposes all this stuff. It could even be in the root URL endpoint like we do in Kinto with capabilities.

Newsletters mapping and acoustic fields are not entirely separate concerns. In the end they both control the columns that will be synchronized to Acoustic.

I'm happy to change if that feels cleaner, no worries.  What would you prefer between:
1. `/configuration`
2. `/acoustic_configuration`
3. `/acoustic_sync_configuration` :) 
4. `/acoustic_fields`, `/acoustic_newsletters_mapping`

leplatrem:
> If it got to this point, they would feel more like separate concerns. 

Oh, I totally forgot about that possible destination for these endpoints. 

As you say, we're not there yet and this endpoint is mostly informational for now. In case one day we introduce a way to manage these lists and mappings from a REST API, we can add new dedicated endpoints (under `/admin/` or whatever).

I will rename it to `/acoustic_configuration`

Issue #469, opened 2022-11-25
Waitlists updated to relationship table over columns


leplatrem:
- [x] Write an ADR
- [ ] Add relationship in schema
- [ ] Deploy/Migrate Data
- [ ] Remove Columns
- [ ] Deploy

PR #470, opened 2022-11-25
Ref #469: waitlists as generic relationship

leplatrem:
Ref #469

### DIFF
--- a/docs/adrs/waitlists_relationships.md
+++ b/docs/adrs/waitlists_relationships.md
@@ -1,0 +1,309 @@
+# Waitlists as Relationships
+
+* Status: proposed
+* Deciders: <CTMS stakeholders> + <Basket stakeholders>
+* Date: November 25, 2022
+
+## Context and Problem Statement
+
+Onboarding new waiting lists requires changing, releasing, and redeploying both CTMS and Basket.
+
+Each waitlist has its own schema with specific collected fields (eg. `country`, `platform`, ...).
+
+Basket uses [some specific code](https://github.com/mozmeao/basket/blob/341facbb2b199bfe2f26488942d0fa251010c1c8/basket/news/backends/ctms.py#L84-L87) to transform flat form data into CTMS nested objects, that has to be modified when a new waitlist has to be sent to CTMS.
+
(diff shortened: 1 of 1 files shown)


bsieber-mozilla:
Thanks for sharing this document @leplatrem; I like how options 2+ reduced engineering and coordination efforts--I believe that this would be an iterative transition from Option 2 to Option 4. 

Since CTMS will be introducing more administration and operational components (as data) to the database, I'm curious what backup processes are like--in a worse case scenario, can we rebuild the database and continue without much stalling?

_Currently waitlist changes are models in the code and are added to the changelog so the DB can be rebuilt with Alembic/SqlAlchemy--do we still have/want/need this functionality?_

leplatrem:
> in a worse case scenario, can we rebuild the database and continue without much stalling?

Well, that's a good question. But I think this is relatively orthogonal to the waitlist issue.
Database backups are necessary for contact data anyway.

Reading between the lines, there could be the question of whether we should use manifests (eg. yaml files) instead of management commands. 

In Remote Settings, we have yaml files that diff/overwrite the metadata in the dev/stage/prod database. If consumers want to update a collection schema, they open a pull-request and when it gets merged, a job takes care of calling the appropriate endpoints to update the metadata. 
This is something we can consider once we have the waitlists code and managements commands in place.

pmac:
I haven't fully digested the doc yet, but this feels at first blush like a step on the way to making newsletters also no longer require code changes. Is that true? Basket treats waiting lists almost exactly the same as newsletters except when the former wants some special field. I'm guessing that country and platform might not be an exhaustive list of possible special fields for future waiting lists either. It would be nice if we could handle it all via data rather than code though.

grahamalama:
> I believe that this would be an iterative transition from Option 2 to Option 4

Agree with @bsieber-mozilla here. An iterative transition from 2 to 4 makes sense to me.

I bet we could get to option 3 pretty quickly after option 2 is implemented, though. We only support 4 waitlists that require transformations at this time. Basket could continue to do explicit transformations for these existing waitlists, but we could support new waitlists with the 
```
{
  "name": "slug", "fields": {...},
}
```
form.

grahamalama (commented):
One additional thing that might be important to document is the role the `newsletters` table will play (or won't play) in this new waitlist approach. Signing up for a waitlist _also_ adds a row to the newsletters table.

```mermaid
erDiagram
    EmailContact }o--o{ newsletters:""
    EmailContact }o--o| RelayWaitlist:""
    EmailContact {
        string email_id PK
    }
    RelayWaitlist {
        string email_id FK
        string geo
        timestamp create_timestamp
        timestamp update_timestamp
    }
    newsletters {
        string email_id FK
        string name "e.g. relay-waitlist"
        boolean subscribed
        timestamp create_timestamp
        timestamp update_timestamp
    }
```

Do we want to have all waitlist data in just the `waitlists` table as a result of this proposed change to the system? If that's the case, we may want to document the removal of waitlist data from the newsletters table. This might also make the migration of the Graceland waitlist tricky since as mentioned in my other comment, there's nothing to indicate that it's a waitlist other than the fact the name of the newsletter slug is `graceland-waitlist`.

leplatrem:
I took the comments of this PR into account in the #492 pull-request.

Closing in favor of #492

PR #475, opened 2022-11-28
Rework linting and testing locally and in CI

grahamalama:
Addresses #462

This PR:
- moves more config to pyproject.toml (and removes unneeded config)
- runs all linting checks through `lint.sh`
- removes extraneous Docker setup
  - consolidates compose files
  - Removes separate linting and testing stages
- reworks our CI workflow
- makes a few other small tweaks

PR #477, opened 2022-11-29
Follow #475: update secrets baseline

leplatrem:
Follow up of #475 

Since Docker on my machine is insupportably painful (like right now I run `make lint` and I get some errors like `docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))` while I just want to LINT MY LOCAL FILES!!!)

I tried to run linting with:

```
poetry install
poetry shell
./docker/lint.sh
```
> which would be my preferred way for `make lint`, but out of scope here

There are errors about secrets. So I ran:

```
./scripts/update_baseline.sh
git add .secrets.baseline
```

But then about `isort` isn't happy about the formatting of imports/wrapping.
Which are fixed by specifying the config file (`profile = "black"`).

Why don't we have these errors when running `./docker/lint.sh` from the container ???

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,5 @@ build
 reports/
 .coverage
 htmlcov/
-.isort.cfg
 docs/_build
 .env
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,20 +1,18 @@
 {
-  "custom_plugin_paths": [],
-  "exclude": {
-    "files": "(poetry.lock$)|(htmlcov/)",
-    "lines": null
-  },
-  "generated_at": "2021-12-07T21:46:42Z",
+  "version": "1.4.0",
   "plugins_used": [
-    {
-      "name": "AWSKeyDetector"
-    },
     {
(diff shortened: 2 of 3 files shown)


grahamalama:
@leplatrem we really shouldn't need that `--settings-path` arg

Based on how you tried to run linting, looks like you ensured you were running the latest version of isort -- I'm assuming that `isort --version` shows `5.10`?

Also, based on how isort [picks up config files](https://pycqa.github.io/isort/docs/configuration/config_files), is there maybe a hidden `.isort.cfg` hanging around in your directory?

leplatrem:
> Also, based on how isort [picks up config files](https://pycqa.github.io/isort/docs/configuration/config_files), is there maybe a hidden .isort.cfg hanging around in your directory?

That was it! 💯

grahamalama (approved):

grahamalama:
@leplatrem could we rename this PR to reflect that what we ended up doing was updating `secrets.baseline`?

Issue #462, opened 2022-11-18
Improve Docker testing / linting setup


grahamalama:
From discussion in #460:

- Investigate why test / lint stages are run in some Docker contexts but not others 
- remove duplicated lint commands between lint.sh and Dockerfile
- lint.sh is not used anywhere -- either use it, or delete it
- a dedicated dockerflow file for lint?

grahamalama:
@leplatrem in #477, you raised the suggestion of running linting locally without Docker.

It's my preference to run linting and testing outside of Docker as well, but I didn't want to make such a change yet until I had buy in from you and @bsieber-mozilla.

How do y'all feel about having Docker focus on just building our production container? That would also have the benefit of simplifying our Dockerfile.

grahamalama:
Okay, I think with #479 and #508, this is complete

PR #479, opened 2022-11-29
Run linting outside Docker

grahamalama:
Addresses #462


bsieber-mozilla (commented):
This is looking good to me! Moving linting checks outside of docker is a great proposition!

leplatrem (approved):
Thanks! 

That's a lot more convenient without Docker. 

I faced this when running `make lint`:

```
Installing dependencies from lock file

No dependencies to install or update
touch .venv/.install.stamp
touch: .venv/.install.stamp: No such file or directory
make: *** [.venv/.install.stamp] Error 1
```

That's because I had a virtualenv already in use. Deleting it solved the issue:

```
poetry env list
poetry env remove ctms-GGbMHGG7-py3.10
```
```

PR #497, opened 2022-12-22
feature(acoustic): add sync loop duration metric

ahoneiser:


ctms/bin/acoustic_sync.py
```diff
@@ -48,21 +48,28 @@ def main(db, settings):
...
             context["trivial"] = True
+
         logger.info(
             "sync_service cycle complete",
             loop_duration_s=round(duration_s, 3),
             loop_sleep_s=round(to_sleep, 3),
             **context
         )
+
+        metric_service.observe_acoustic_sync_loop_duration(round(duration_s, 3))
```
grahamalama, line 68: Oh, so we're tracking how long a batch takes to sync before we sleep?

grahamalama (commented):
Could we add some context here (in the commit and/or the PR description) about what metric(s) we're adding and why?

ctms/bin/acoustic_sync.py
ahoneiser: right, should tell us how to schedule the healthcheck intervals

grahamalama (approved):
Lgtm 🚢

PR #500, opened 2023-01-05
Remove all bigquery import machinery

grahamalama:
This code was written to import contact data from bigquery into CTMS. We no longer perform the operations contained in these scripts, so they can be safely deleted.


leplatrem:
I believe you could also remove the `*TableSchema`, like https://github.com/mozilla-it/ctms-api/blob/0c573e900776764edde972af5cb8e624a87a1a2a/ctms/schemas/newsletter.py#L58-L73

grahamalama:
Ahh nice, good catch @leplatrem.

I'm going to merge #501, then remove `google-cloud-bigquery` in this PR, then consider it ready for review.

bsieber-mozilla (approved):
:shipit: 

🧹 thanks for this cleanup work @grahamalama

PR #503, opened 2023-01-06
Always run container with single Uvicorn process

### LINKED ISSUE #494: Run app as a single process in production
See also: https://github.com/mozilla/jira-bugzilla-integration/issues/133

This is what the [FastAPI docs](https://fastapi.tiangolo.com/deployment/docker/?h=deployment#replication-number-of-processes) recommend:
> If you have a cluster of machines with Kubernetes ... then you will probably want to handle replication at the cluster level instead of using a process manager (like Gunicorn with workers) in each container.

This could also help us instill confidence in our Prometheus metrics (https://github.com/mozilla-it/ctms-api/issues/447). See [their caveats](https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn) for collecting metrics across processes.

grahamalama:
In this PR, we run remove Gunicorn and run the container with a single process of Unvicorn in both development and production. See #494 for rationale of why we're making this change.

With removing Gunicorn, we also had to do some tweaking to our logging and application settings.

Closes #494


bkochendorfer (approved):
Seems cool, once this is in dev let's see if we need to change the minimum number of pods in your deployment to make up for the change here.

PR #508, opened 2023-01-10
Move testing outside Docker, simplify Dockerfile

grahamalama:
Addresses #462

In this PR, we move testing outside of Docker. Since the Dockerfile is now only responsible for building a production container, we also simplify the Dockerfile accordingly.


grahamalama:
TODO: do a once-over of our documentation and update what's now outdated

bin/update_and_install_system_packages.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
grahamalama, line 1: Adapted from [this script](https://pythonspeed.com/articles/system-packages-docker/#debian-ubuntu-and-the-debian-based-python-base-image)

Makefile
```diff
@@ -17,14 +17,14 @@ help:
...
-	@echo "  shell       - open a shell in the web container"
-	@echo "  db-only     - run PostgreSQL server"
+	@echo "  build          - build docker containers"
+	@echo "  db-only        - run PostgreSQL server"
+	@echo "  lint           - lint check for code"
+	@echo "  setup          - (re)create the database"
+	@echo "  shell          - open a shell in the web container"
+	@echo "  start          - run the API service"
+	@echo "  test           - run test suite"
+	@echo "  update-secrets - scan repository for secrets and update baseline file, if necessary"
```
bsieber-mozilla, line 27: _thanks for this addition!_

bin/update_baseline.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -euo pipefail
+
+detect-secrets scan . \
+--baseline .secrets.baseline \
+--exclude-files "(poetry.lock$)|(htmlcov/)"
```
bsieber-mozilla, line 7: thanks for aggregating these all under /bin! (and the readme update!)

bsieber-mozilla (approved):
Woah! Docker is way simpler with this change--thanks @grahamalama !

grahamalama:
Thanks for the review @bsieber-mozilla! 

@bkochendorfer, care to give this a once-over before we ship this to staging since this is a pretty big change to how we build the container?

.github/workflows/test-build.yaml
```diff
@@ -23,25 +23,17 @@ jobs:
...
+        run: pipx install poetry
+      - uses: actions/setup-python@v4
         with:
-          context: .
-          file: docker/Dockerfile
-          push: false
-          target: "development"
-          tags: ghcr.io/${{ github.repository }}:${{ github.sha }}
-
+          python-version: "3.10"
```
bsieber-mozilla, line 31: will dependabot notice and be able to update this python-version; or is that something for us to keep in mind?

docker/Dockerfile
```diff
@@ -1,91 +1,44 @@
...
-      build-essential \
-      libpq-dev
+    POETRY_VIRTUALENVS_IN_PROJECT=true \
+    PYSETUP_PATH="/opt/pysetup"
 
-# Install Poetry - respects $POETRY_VERSION & $POETRY_HOME
-USER app
 RUN python3 -m venv $POETRY_HOME && \
 $POETRY_HOME/bin/pip install poetry==1.3.1 && \
 $POETRY_HOME/bin/poetry --version
```
bsieber-mozilla, line 13: I wonder...

`poetry==1.3.1`
maybe instead we could move this to be an env var above:
```
...
POETRY_HOME=/opt/poetry\
POETRY_VERSION=1.3.1
... ```

I was looking at this reference from the [poetry installation guide](https://python-poetry.org/docs/1.2#installing-with-the-official-installer)

**Review by bkochendorfer — APPROVED** [review]
This looks good to me. Is there a reason you wanted to move the scripts into bin?

**Comment by grahamalama** [comment]
> Is there a reason you wanted to move the scripts into bin?

@bkochendorfer for moving scripts, I just wanted to consolidate all scripts into one place. In fact, in a future PR, I _might_ move scripts from `ctms/bin` to the top level `bin`, since I think that would let us leave the `PYTHONPATH` alone.

As for why I chose `bin`, I think that was just muscle memory from previous projects.

File: docker/Dockerfile (hunk 5 above)
Comment (grahamalama) [reply docker/Dockerfile:13]: @bsieber-mozilla I think this is more relevant when we're using the installation script `curl`ed from `https://install.python-poetry.org`.

For the pip install version (like what we're doing), from this snippet in the docs, it looks like they inline the poetry version:
```
export POETRY_HOME=/opt/poetry
python3 -m venv $POETRY_HOME
$POETRY_HOME/bin/pip install poetry==1.2.0
$POETRY_HOME/bin/poetry --version
```

We could move that version to a variable, but we'd only end up using that number here.

.github/workflows/test-build.yaml
grahamalama: I think this is something we'll just have to remember, unfortunately. But it would definitely be nice if we could keep these things in sync. Maybe there's a way 🤔

PR #510, opened 2023-01-13
Streamline environment bootstrapping, move contents of `docker/` to top level

grahamalama:
In this PR, we:
- consolidate sources of local development config (`.env.dist` `local_dev.env`) to one `.env.example`, which is copied to `.env` with `Make`
- move contents of `docker/` to top level, which allows us to remove some config options with Docker commands
- delete `docker/log_config.json`, which appeared to be unused


bsieber-mozilla (approved):
Nice! Thanks for organizing and cleaning this code up! 
🚢

Makefile
```diff
@@ -32,8 +32,8 @@ help:
 .env:
 	@if [ ! -f .env ]; \
 	then \
-	echo "Copying env.dist to .env..."; \
-	cp docker/config/env.dist .env; \
+	echo "Copying .env.example to .env"; \
+	cp .env.example .env; \
```
leplatrem, line 36: I don't see why this has to be so complicated. 
The rule isn't executed if `.env` exists...

Example:
https://github.com/mozilla-services/telescope/blob/f4739c65d7309ef39bdebd3c198e618ade89fc36/Makefile#L44-L45

grahamalama: Good catch. I was mostly just find/replacing, but good opportunity to fix this

PR #520, opened 2023-01-23
Remove SQL assertions

grahamalama:
In this PR, we:
- Remove test assertions that check for how many queries happened during a function invocation
- Remove assertions about the textual content of SQL statements generated by the ORM
- Remove the `StatementWatcher` class that allowed us to track these things

These assertions have seemed to make tests brittle and test implementation details. 

If it _is_ important to verify either the number of queries that a particular function emits or the SQL content of those queries, we should find a way to incorporate that into the test suite in a way that doesn't require changing `assert watcher.count == n` for every test function.

### DIFF
--- a/tests/unit/test_crud.py
+++ b/tests/unit/test_crud.py
@@ -5,7 +5,6 @@ from typing import List
 from uuid import uuid4
 
 import pytest
-import sqlalchemy.event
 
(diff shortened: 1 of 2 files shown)


leplatrem (approved):
👍

PR #532, opened 2023-02-02
Bump Python to 3.11.1

grahamalama:
This PR bumps the version of Python we use to run CTMS from `3.10.9` to `3.11.1`.

In order to complete the upgrade, we also had to:
- update `pylint`, and make some related changes
- re-add `uvicorn` so we pulled in `uvloop>=0.17`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ authors = ["CTMS Reviewers <@mozilla-it/ctms-reviewers>"]
 
 [tool.poetry.dependencies]
 # These packages are mandatory and form the core of this package’s distribution.
-python = "3.10.9"
+python = "3.11.1"
 fastapi = "^0.84.0"
 starlette = "^0.19.1"
 requests = "^2.28.1"
@@ -46,7 +46,7 @@ types-python-dateutil = "^2.8.19"
 detect-secrets = "^1.4.0"
 bandit = "^1.7.0"
 SQLAlchemy-Utils = "^0.39.0"
-pylint = "^2.13.9"
+pylint = "^2.16.0"
 pylint-pytest = "^1.1.2"
 types-requests = "^2.28.11"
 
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -58,18 +58,13 @@ testpaths = [
 [tool.pylint]
     [tool.pylint.'MESSAGES CONTROL']
     disable = [
-        # https://black.readthedocs.io/en/stable/compatible_configs.html
-        "C0330",
```
grahamalama, line 62: When upgrading Pylint, it emitted warnings that these config options are no longer supported

pyproject.toml
```diff
@@ -58,18 +58,13 @@ testpaths = [
...
-        "C0326",
-
         "R0801", # sqla has a lot of duplicate code and it is work to "fix" this
         "C0301", # if black is happy with the line-length, let's not mess with it
         "R0903", # this is triggered by pydantic's config and is really just a taste thing anyway
         "R0913", # we have too-many-arguments in too many places to easily fix
         "R0912", # Same idea as above, let's leave it since it is a matter of taste
         "W0511", # Let us have our TODOs
         "W0613", # sometimes unused arguments are nice for symmetry and also api methods
-        "R0201", # Not using "self" when it is defined is a matter of taste
```
grahamalama, line 72: This config option is also no longer supported

pyproject.toml
```diff
@@ -120,7 +115,7 @@ profile = "black"
 
 [tool.mypy]
 # Sync Python version with Dockerfile
-python_version = "3.10.9"
```
grahamalama, line 123: This config option only accepts specifying Python with a major and minor version -- it does not accept adding the patch version

bsieber-mozilla (approved):
Thanks for these changes @grahamalama!

----

Not a task for this PR but...
I'm curious if we can have a "mark/#todo" where all these versions of python are--mainly hoping to make it easier to track this kind of update (or automate it). I think in looking to automate it, there might be some "chicken and the egg" type problems on where the "source" might be. (I _think_ pyproject.toml could be the "source" and through TOML-parsing, we apply the correct version to the other files? 🤷 )

grahamalama:
@bsieber-mozilla yeah, it's not perfect. Usually a find + replace works for me -- `3.11`. It's a pretty manual process, but probably doesn't happen often enough to warrant a thorough automated process. We did add [this section](https://github.com/mozilla-it/ctms-api/blob/main/docs/developer_setup.md#python) to the docs recently, which kind of gets at this.

PR #538, opened 2023-02-06
Run formatting black 23.1.0

leplatrem:
Apparently my usage of the "auto-merge" feature isn't appropriate (yet?).

See #537 , that was merged with failing `lint` CI test 🤷


grahamalama (commented):
In the `test*` files where we have `(main, _, _,)`,could we delete the trailing comma so black doesn't split that out onto multiple lines?

grahamalama (approved):
One more with `(main, newsletters, _,)`, then this is good to go!

Issue #52, opened 2021-02-24
Setup cascades for secondary data tables


jwhitlock:
The secondary tables, such as ``newsletter`` and ``amo`` should stay in sync with the primary ``email`` table. 

At the SQL level, this implies ``ON DELETE CASCADE`` clauses on the ``FOREIGN KEY`` relationship.

At the SQLAlchemy level, [cascades](https://docs.sqlalchemy.org/en/13/orm/cascades.html#) refers to this and some other ORM-level tracking. I think what we want is ``all, delete-orphan``, but there are lots of options:

* ``save-update`` (default on, set by ``all``) - When an item is inserted or updated, also insert or update related data. Should be enabled.
* ``merge`` (default on, set by ``all``) - When an item is merged into a SQLAlchemy session, also merge relationships. [Session merging](https://docs.sqlalchemy.org/en/13/orm/session_state_management.html#merging) is a SQLAlchemy concept that I hope you never need to understand. This should be enabled.
* ``delete`` (set by ``all``) - When a "parent" is deleted, also delete the "child". This should be enabled for the secondary tables, and not on the email relationship. The section [Using foreign key ON DELETE cascade with ORM relationships](https://docs.sqlalchemy.org/en/13/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships) shows how to set this so that the ``ON DELETE CASCADE`` will be set at the SQL level.
* ``delete-orphan`` - When de-associated with the parent, delete the orphaned child. This should be set on the child tables
* ``refresh-expire`` (set by ``all``) - When a parent is passed to [``Sesion.expire()``](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.expire), also expire children. I need to understand ``Session`` more to fully appreciate this one, but setting it seems like the safe thing.
* ``expunge`` (set by ``all``) - When a parent is removed with [``Session.expunge()``](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.expunge), then also expunge the children. Setting it seems like the safe thing.

I think this will need a migration to set the ``ON DELETE CASCADE`` on the foreign keys.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-53)

PR #548, opened 2023-02-14
Fix #240, #529: Add GDPR deletion endpoint

### LINKED ISSUE #240: Make GDPR requests easier
For GDPR (et al.) we'll need to provide all the info we have on a given person (email address) and the ability to delete said information.

I imagine there are several ways to accomplish this:
1. Do it all through the CTMS API (explicitly in python?). Probably the Right Way to do this?
2. Do it through CTMS API but using postgres functions
3. Do it directly in the DB using pre-written functions
4. Do it directly in the DB table-by-table

For deletes, make things a bit easier (arguably) by making all the sub-tables ON DELETE CASCADE from `emails`?

I've written basically the same postgres function twice (it'd be easy to combine into one function if desired), once for lookups and once for deletes that can be used expicitly, or as a starting point off which to base other work. The problem with using a postgres function is, of course, that it's somewhat hidden. Anyway, here they are:
```
 CREATE OR REPLACE FUNCTION gdpr_lookup(email_to_find text) RETURNS boolean AS $$
  DECLARE
    email_ids record;
    table_name text;
    --TODO: make these table names an external_table? Like `gdpr_relevant_tables` or something?
    table_names text[] := ARRAY['amo', 'fxa', 'mofo', 'newsletters', 'vpn_waitlist', 'emails'];
    results record;
    result boolean;
  BEGIN
    result := false;
    FOR email_ids IN
      SELECT email_id,'emails' AS lookup_table_name FROM emails WHERE LOWER(primary_email)=LOWER(email_to_find)
      UNION ALL
      SELECT email_id,'fxa' AS lookup_table_name FROM fxa WHERE LOWER(primary_email)=LOWER(email_to_find)
    LOOP
      CONTINUE WHEN email_ids.email_id IS NULL;
      RAISE NOTICE 'Email % (%) found in primary_email field of table %', email_to_find, email_ids.email_id, email_ids.lookup_table_name;
      result := true;
      FOREACH table_name IN ARRAY table_names
      LOOP
        RAISE NOTICE 'Searching % for %', table_name, email_ids.email_id;
        EXECUTE format('SELECT * FROM %I WHERE email_id=$1', table_name)
          INTO results
          USING email_ids.email_id;
        RAISE NOTICE '%', results;
      END LOOP;
    END LOOP;
    RETURN result;
  END;
$$ LANGUAGE plpgsql;
```
and
```
CREATE OR REPLACE FUNCTION gdpr_delete(email_to_find text) RETURNS boolean AS $$
  DECLARE
    email_ids record;
    table_name text;
    --TODO: make these table names an external_table? Like `gdpr_relevant_tables` or something?
    table_names text[] := ARRAY['amo', 'fxa', 'mofo', 'newsletters', 'vpn_waitlist', 'emails'];
    results record;
    result boolean;
  BEGIN
    result := false;
    FOR email_ids IN
      SELECT email_id,'emails' AS lookup_table_name FROM emails WHERE LOWER(primary_email)=LOWER(email_to_find)
      UNION ALL
      SELECT email_id,'fxa' AS lookup_table_name FROM fxa WHERE LOWER(primary_email)=LOWER(email_to_find)
    LOOP
      CONTINUE WHEN email_ids.email_id IS NULL;
      RAISE NOTICE 'Email % (%) found in primary_email field of table %', email_to_find, email_ids.email_id, email_ids.lookup_table_name;
      result := true;
      FOREACH table_name IN ARRAY table_names
      LOOP
        RAISE NOTICE 'Deleting % from %', email_to_find, table_name;
        EXECUTE format('DELETE FROM %I WHERE email_id=$1 RETURNING *', table_name)
          INTO results
          USING email_ids.email_id;
        RAISE NOTICE '%', results;
      END LOOP;
    END LOOP;
    RETURN result;
  END;
$$ LANGUAGE plpgsql;
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-24)

### LINKED ISSUE #529: improve DSARS/GDPR deletion request handling
# tracking issue for DSARS/GDPR deletion requests

DSARS/GDPR deletion requests are currently handled in a primarily manual process, a couple of the tasks could be automated to reduce maintenance of resources as well as to reduce SRE time. Ideas to improve the process:

* add an API endpoint to CTMS to make the workflow easier to handle
* add a script / k8s cronjob to do the actual deletion from within the application pods

ahoneiser:
Fix #240 
Fix #529

### DIFF
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -39,7 +39,7 @@ from .config import re_trace_email
 from .crud import (
     create_contact,
     create_or_update_contact,
-    gdpr_delete,
+    delete_contact,
     get_all_acoustic_fields,
     get_all_acoustic_newsletters_mapping,
     get_api_client_by_id,
@@ -76,7 +76,6 @@ from .schemas import (
     CTMSSingleResponse,
     EmailSchema,
     FirefoxAccountsSchema,
-    GDPRDeleteResponse,
     IdentityResponse,
     MozillaFoundationSchema,
     NotFoundResponse,
@@ -692,6 +691,32 @@ def partial_update_ctms_contact(
     return get_ctms_response_or_404(db=db, email_id=email_id)
 
 
+@app.delete(
+    "/ctms/{primary_email}",
+    summary="Delete all contact information from primary email",
+    response_model=List[IdentityResponse],
+    responses={
+        404: {"model": NotFoundResponse},
+    },
+    tags=["Public"],
+)
+def delete_contact_by_primary_email(
+    primary_email: str,
+    db: Session = Depends(get_db),
+    api_client: ApiClientSchema = Depends(get_enabled_api_client),
+):
+    ids = all_ids(primary_email=primary_email.lower())
+    contacts = get_contacts_by_ids(db, **ids)
+
+    if not contacts:
+        raise HTTPException(status_code=404, detail=f"email {primary_email} not found!")
+
+    for contact in contacts:
+        delete_contact(db=db, email_id=contact.email.email_id)
+
+    return [contact.as_identity_response() for contact in contacts]
+
+
 @app.get(
     "/updates",
     summary="Get all contacts within provided timeframe",
@@ -1122,42 +1147,6 @@ def stripe_pubsub(
         }
         return JSONResponse(content=content, status_code=202)
     return {"status": "OK", "count": count}
-
-
-@app.delete(
-    "/ctms/gdpr/{email}",
-    tags=["Public"],
-    summary="delete email from database",
-    responses={
-        404: {"model": NotFoundResponse},
-    },
-    response_model=GDPRDeleteResponse,
-)
-def gdpr_delete_email(
-    email: str,
-    db: Session = Depends(get_db),
-    api_client: ApiClientSchema = Depends(get_enabled_api_client),
-):
-    email_to_find = email.lower()
-    ids = all_ids(primary_email=email_to_find)
-    contacts = get_contacts_by_ids(db, **ids)
-
-    if not contacts:
-        raise HTTPException(status_code=404, detail=f"email {email} not found!")
-
-    dropped = list()
-
-    for contact in contacts:
-        gdpr_delete(db=db, email_id=contact.email.email_id)
-
-        dropped.append(
-            {
-                "primary_email": contact.email.primary_email,
-                "email_id": str(contact.email.email_id),
-            }
-        )
-
-    return GDPRDeleteResponse(status="ok", dropped=dropped)
 
 
 if __name__ == "__main__":
--- a/ctms/app.py
+++ b/ctms/app.py
@@ -39,6 +39,7 @@ from .config import re_trace_email
 from .crud import (
     create_contact,
     create_or_update_contact,
+    gdpr_delete,
     get_all_acoustic_fields,
     get_all_acoustic_newsletters_mapping,
     get_api_client_by_id,
@@ -75,6 +76,7 @@ from .schemas import (
     CTMSSingleResponse,
     EmailSchema,
     FirefoxAccountsSchema,
+    GDPRDeleteResponse,
     IdentityResponse,
     MozillaFoundationSchema,
     NotFoundResponse,
@@ -1120,6 +1122,42 @@ def stripe_pubsub(
         }
         return JSONResponse(content=content, status_code=202)
     return {"status": "OK", "count": count}
+
+
+@app.delete(
+    "/ctms/gdpr/{email}",
+    tags=["Public"],
+    summary="delete email from database",
+    responses={
+        404: {"model": NotFoundResponse},
+    },
+    response_model=GDPRDeleteResponse,
+)
+def gdpr_delete_email(
+    email: str,
+    db: Session = Depends(get_db),
+    api_client: ApiClientSchema = Depends(get_enabled_api_client),
+):
+    email_to_find = email.lower()
+    ids = all_ids(primary_email=email_to_find)
+    contacts = get_contacts_by_ids(db, **ids)
+
+    if not contacts:
+        raise HTTPException(status_code=404, detail=f"email {email} not found!")
+
+    dropped = list()
+
+    for contact in contacts:
+        gdpr_delete(db=db, email_id=contact.email.email_id)
+
+        dropped.append(
+            {
+                "primary_email": contact.email.primary_email,
+                "email_id": str(contact.email.email_id),
+            }
+        )
+
+    return GDPRDeleteResponse(status="ok", dropped=dropped)
 
 
 if __name__ == "__main__":
--- a/ctms/bin/gdpr
+++ b/ctms/bin/gdpr
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+import sys
+
+import click
+from pydantic import BaseSettings
+from requests import Session
+from requests.auth import HTTPBasicAuth
+
+
+class Settings(BaseSettings):
+    api_url: str = "http://127.0.0.1:8000"
+    client_id: str
+    client_secret: str
+    session: Session
+
+    class Config:
+        env_prefix = "ctms_"
+
+
+@click.group()
+@click.pass_context
+def cli(ctx):
+    """ctms cli for gdpr actions."""
+    ctx.obj = Settings(session=Session())
+
+    response = ctx.obj.session.post(
+        f"{ctx.obj.api_url}/token",
+        auth=HTTPBasicAuth(ctx.obj.client_id, ctx.obj.client_secret),
+    )
+
+    if not response.ok:
+        print(f"{response.status_code}: {response.reason}", file=sys.stderr)
+
+    else:
+        data = response.json()
+        token = data["access_token"]
+
+        ctx.obj.session.headers.update({"Authorization": f"Bearer {token}"})
+
+
+@cli.command()
+@click.option("--email")
+@click.option("--email-list", type=click.File("r"))
+@click.pass_obj
+def delete(obj, email, email_list):
+    """delete single contact or a list of contacts from ctms"""
+    to_be_deleted = list()
+
+    if email_list:
+        for line in email_list.readlines():
+            to_be_deleted.append(line.rstrip())
+
+    if email:
+        to_be_deleted.append(email)
+
+    for item in to_be_deleted:
+        response = obj.session.delete(f"{obj.api_url}/ctms/{item}")
+
+        if not response.ok:
+            if response.status_code == 404:
+                print(f"{item} not found in CTMS")
+
+            else:
+                print(f"{response.status_code}: {response.reason}", file=sys.stderr)
+
+        else:
+            data = response.json()
+
+            for contact in data:
+                email_id = contact["email_id"]
+                msg = f"DELETING {item} (ctms id: {email_id})."
+
+                if contact["fxa_id"]:
+                    msg += " fxa: YES."
+
+                if contact["mofo_contact_id"]:
+                    msg += " mofo: YES."
+
+                print(msg)
+
+
+if __name__ == "__main__":
+    cli()
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -929,3 +929,19 @@ def delete_acoustic_newsletters_mapping(dbsession, source):
     dbsession.delete(row)
     dbsession.commit()
     return row
+
+
+def gdpr_delete(db: Session, email_id: str):
+    models = [
+        PendingAcousticRecord,
+        AmoAccount,
+        MozillaFoundationContact,
+        Newsletter,
+        Waitlist,
+        FirefoxAccount,
+        Email,
+    ]
+
+    for model in models:
+        db.query(model).filter(model.email_id == email_id).delete()
+        db.commit()
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -597,6 +597,22 @@ def create_or_update_contact(
     create_or_update_waitlists(db, email_id, contact.waitlists)
 
 
+def delete_contact(db: Session, email_id: UUID4):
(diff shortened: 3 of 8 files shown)


ctms/app.py
```diff
@@ -1122,5 +1124,41 @@ def stripe_pubsub(
     return {"status": "OK", "count": count}
 
 
+@app.delete(
+    "/ctms/gdpr/{email}",
+    tags=["Public"],
+    summary="delete email from database",
```
leplatrem, line 1130: Something like `Delete all contact information from primary email`?

ctms/app.py
```diff
@@ -1122,5 +1124,41 @@ def stripe_pubsub(
...
+        gdpr_delete(db=db, email_id=contact.email.email_id)
+
+        dropped.append(
+            {
+                "primary_email": contact.email.primary_email,
+                "email_id": str(contact.email.email_id),
+            }
+        )
+
+    return GDPRDeleteResponse(status="ok", dropped=dropped)
```
leplatrem, line 1160: We could return the whole contact info like in the `/identities` endpoints.
```py
return [contact.as_identity_response() for contact in contacts]
```

ctms/schemas/__init__.py
```diff
@@ -26,6 +26,7 @@
 from .mofo import MozillaFoundationInSchema, MozillaFoundationSchema
 from .newsletter import NewsletterInSchema, NewsletterSchema, UpdatedNewsletterInSchema
 from .product import ProductBaseSchema
+from .response import GDPRDeleteResponse
```
leplatrem, line 29: We have responses schemas in `schemas/contact.py` (eg. `CTMSSingleResponse`). We could move it to your new module, or move `DeleteResponse` there

ctms/bin/gdpr
```diff
@@ -0,0 +1,93 @@
...
+        f"{ctx.obj.api_url}/token",
+        auth=HTTPBasicAuth(ctx.obj.client_id, ctx.obj.client_secret),
+    )
+
+    if not response.ok:
+        print(f"{response.status_code}: {response.reason}", file=sys.stderr)
+
+    else:
+        data = response.json()
+        token = data.get("access_token")
```
leplatrem, line 37:
```suggestion
        token = data["access_token"]
```

ctms/bin/gdpr
```diff
@@ -0,0 +1,93 @@
...
+
+        ctx.obj.session.headers.update(
+            {"Content-Type": "application/json", "Authorization": f"Bearer {token}"}
```
leplatrem, line 40: Shall we put the `Content-Type` part out of the access token logics?

ctms/bin/gdpr
```diff
@@ -0,0 +1,93 @@
...
+            f"{obj.api_url}/ctms", params={"primary_email": item}
+        )
+
+        if not response.ok:
+            print(f"{response.status_code}: {response.reason}", file=sys.stderr)
+
+        else:
+            data = response.json()
+
+            if len(data):
```
leplatrem, line 70: I don't think this step is necessary, since the endpoint already returns a 404 when the contact is not found.

If you want to check before delete, you could introduce a `--dry-run` flag for example, and do a get() instead of a delete() when set

ctms/crud.py
```diff
@@ -929,3 +929,19 @@ def delete_acoustic_newsletters_mapping(dbsession, source):
...
+def gdpr_delete(db: Session, email_id: UUID4):
+    models = [
+        PendingAcousticRecord,
+        AmoAccount,
+        MozillaFoundationContact,
+        Newsletter,
+        Waitlist,
+        FirefoxAccount,
+        Email,
+    ]
```
leplatrem: I wonder if we should add `ondelete='CASCADE'` to our `db.ForeignKey()` relatioships instead of doing it manually here at the Python level.

Ideally, deleting a `Email` record, should delete everything related to it, no?

ctms/crud.py
```diff
@@ -929,3 +929,19 @@ def delete_acoustic_newsletters_mapping(dbsession, source):
     dbsession.delete(row)
     dbsession.commit()
     return row
+
+
+def gdpr_delete(db: Session, email_id: UUID4):
```
leplatrem, line 934: ditto about `gdpr` here

leplatrem (commented):
Thanks Andre for tackling this!

You've got a real good base here, my feedback is mostly about API design and naming. Feel free to reach out if something doesn't look to relevant

ctms/app.py
```diff
@@ -1122,5 +1124,41 @@ def stripe_pubsub(
     return {"status": "OK", "count": count}
 
 
+@app.delete(
+    "/ctms/gdpr/{email}",
```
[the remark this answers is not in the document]
ahoneiser: agree, changed it :slightly_smiling_face:

ctms/app.py
ahoneiser: updated the summary :+1:

ctms/app.py
ahoneiser: makes way more sense to me, dropped the GDPRDeleteResponse :+1:

ctms/bin/gdpr
ahoneiser: true, removed that part :slightly_smiling_face:

ctms/crud.py
ahoneiser: right, this could be handled through the database model :thinking:

cvalaas (approved):

ctms/app.py
```diff
@@ -690,6 +691,41 @@ def partial_update_ctms_contact(
...
+
+    for contact in contacts:
+        delete_contact(db=db, email_id=contact.email.email_id)
+
+        dropped.append(
+            {
+                "primary_email": contact.email.primary_email,
+                "email_id": str(contact.email.email_id),
+            }
+        )
```
leplatrem, line 724: Here `dropped` is not used anymore
You could log these infos though

ctms/schemas/response.py
```diff
@@ -0,0 +1,5 @@
+from pydantic import BaseModel
+
+
+class Response(BaseModel):
+    status: str
```
leplatrem, line 5: This isn't used anymore :)

ctms/bin/gdpr
```diff
@@ -0,0 +1,85 @@
...
+        if not response.ok:
+            if response.status_code == 404:
+                print(f"{item} not found in CTMS")
+
+            else:
+                print(f"{response.status_code}: {response.reason}", file=sys.stderr)
+
+        else:
+            data = response.json()
+            contact = data[0]
```
leplatrem, line 71: Why only considering the first contact? Maybe you could print out how many contacts matched the email at least?

ctms/bin/gdpr
```diff
@@ -0,0 +1,85 @@
...
+            email_id = contact["email_id"]
+            msg = f"DELETING {item} (ctms id: {email_id})."
```
leplatrem, line 73: nitpick: this happened already, so `DELETED` instead of `DELETING`

ctms/bin/gdpr
```diff
@@ -0,0 +1,85 @@
...
+
+    for item in to_be_deleted:
+        response = obj.session.delete(f"{obj.api_url}/ctms/{item}")
+
+        if not response.ok:
+            if response.status_code == 404:
+                print(f"{item} not found in CTMS")
+
+            else:
+                print(f"{response.status_code}: {response.reason}", file=sys.stderr)
```
leplatrem, line 67: If there was an unexpected error like a 5XX, do you want to proceed with other emails or give up?

leplatrem (commented):
You'll have to add a test. Using the `maximal_contact` test fixture, you should cover most related tables. You'd have to test the `404` response too

ctms/schemas/response.py
ahoneiser: right, getting rid of that :)

ctms/bin/gdpr
ahoneiser: right .. output of that script is going to be added to a Spreadsheet that has conditional formatting in place .. I'm afraid this needs to be kept until another process gets evolved around that Spreadsheet :(

ctms/bin/gdpr
ahoneiser: proceeding is okay for me, unexpected errors are logged, so they can be handled .. not seeing a benefit for either way atm :thinking:

ctms/bin/gdpr
ahoneiser: thought of the primary_email to be a unique property, so took the first item from the response .. will put a loop around this :+1:

ctms/app.py
ahoneiser: right, getting rid of it :+1:

tests/unit/test_api_delete.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_delete_contact_by_primary_email_not_found(client):
+    resp = client.delete("/ctms/foo@bar.com")
+    assert resp.status_code == 404
+
+
+def test_delete_contact_by_primary_email(client, minimal_contact):
+    primary_email = minimal_contact.email.primary_email
+    resp = client.delete(f"/ctms/{primary_email}")
+    assert resp.status_code == 200
```
leplatrem, line 12: Could you check that delete actually deleted something? For example, that a subsequent get or delete would return 404?

leplatrem: Also, using `maximal_contact` would allow to test more of the model deletions (since it has more related objects)

ctms/crud.py
```diff
@@ -597,6 +597,22 @@ def create_or_update_contact(
...
+        PendingAcousticRecord.email_id == email_id
+    ).delete()
+    db.query(AmoAccount).filter(AmoAccount.email_id == email_id).delete()
+    db.query(MozillaFoundationContact).filter(
+        MozillaFoundationContact.email_id == email_id
+    ).delete()
+    db.query(Newsletter).filter(Newsletter.email_id == email_id).delete()
+    db.query(Waitlist).filter(Waitlist.email_id == email_id).delete()
+    db.query(FirefoxAccount).filter(FirefoxAccount.email_id == email_id).delete()
+    db.query(Email).filter(Email.email_id == email_id).delete()
```
leplatrem, line 611: Can you try to add `cascade="save-update, merge, delete",` on the `Email` model relationships and see if you obtain the same result by then just deleting the `Email` objects?

```diff
diff --git a/ctms/models.py b/ctms/models.py
index 8b48e3e..35284f7 100644
--- a/ctms/models.py
+++ b/ctms/models.py
@@ -57,15 +57,34 @@ class Email(Base):
     )

     newsletters = relationship(
-        "Newsletter", back_populates="email", order_by="Newsletter.name"
+        "Newsletter",
+        back_populates="email",
+        order_by="Newsletter.name",
+        cascade="save-update, merge, delete",
     )
     waitlists = relationship(
-        "Waitlist", back_populates="email", order_by="Waitlist.name"
+        "Waitlist",
+        back_populates="email",
+        order_by="Waitlist.name",
+        cascade="save-update, merge, delete",
+    )
+    fxa = relationship(
+        "FirefoxAccount",
+        back_populates="email",
+        uselist=False,
+        cascade="save-update, merge, delete",
+    )
+    amo = relationship(
+        "AmoAccount",
+        back_populates="email",
+        uselist=False,
+        cascade="save-update, merge, delete",
     )
-    fxa = relationship("FirefoxAccount", back_populates="email", uselist=False)
-    amo = relationship("AmoAccount", back_populates="email", uselist=False)
     mofo = relationship(
-        "MozillaFoundationContact", back_populates="email", uselist=False
-    amo = relationship("AmoAccount", back_populates="email", uselist=False)
     mofo = relationship(
-        "MozillaFoundationContact", back_populates="email", uselist=False
+        "MozillaFoundationContact",
+        back_populates="email",
+        uselist=False,
+        cascade="save-update, merge, delete",
     )
     stripe_customer = relationship(
         "StripeCustomer",
@@ -74,6 +93,12 @@ class Email(Base):
         primaryjoin="Email.email_id==FirefoxAccount.email_id",
         secondaryjoin="remote(FirefoxAccount.fxa_id)==foreign(StripeCustomer.fxa_id)",
         secondary="join(FirefoxAccount, StripeCustomer, FirefoxAccount.fxa_id == StripeCustomer.fxa_id)",
+        cascade="save-update, merge, delete",
+    )
+    acoustic_records = relationship(
+        "PendingAcousticRecord",
+        back_populates="email",
+        cascade="save-update, merge, delete",
     )

File: tests/unit/test_api_delete.py [hunk 17]
```diff
@@ -0,0 +1,12 @@
+"""Unit tests for DELETE /ctms/{primar_email}"""
+
+
+def test_delete_contact_by_primary_email_not_found(client):
+    resp = client.delete("/ctms/foo@bar.com")
+    assert resp.status_code == 404
+
```
leplatrem, line 7: Can you add a test that verifies that an anonymous client cannot delete also please?

ctms/crud.py
leplatrem: I realize we have https://github.com/mozilla-it/ctms-api/issues/52

ahoneiser: #52 resolves this, right?

tests/unit/test_api_delete.py
ahoneiser: done :+1:

ctms/crud.py
leplatrem: yes!

tests/unit/test_api_delete.py
leplatrem: yes, scratch that, sorry!

leplatrem (approved):
Great job Andre! I think this is now good to ship! 🎉  

If there was any further improvement to add, we can still open follow-up, since this new endpoint does not interfere with the rest of the application.

PR #550, opened 2023-02-16
Fix #52: Add delete cascade on Email relationships

### LINKED ISSUE #52: Setup cascades for secondary data tables

leplatrem:
Fix #52

* [ ] Add unit test of cascade
* [ ] Remove manual cascaded deleted introduced in #548 
* [ ] Add cascade option on relationships for the ORM part
* [ ] Add single parent option?


grahamalama:
With a quick look, is there a reason you went with `ondelete` instead of [`cascade`](https://docs.sqlalchemy.org/en/14/orm/cascades.html#cascades)?

grahamalama:
Ahh, maybe we actually need [both things](https://docs.sqlalchemy.org/en/14/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships)

leplatrem:
@grahamalama @bsieber-mozilla I think we can merge this PR as it is, and have a follow-up for the ORM annotation if it seems necessary

bsieber-mozilla (approved):
Thanks for cleaning this up!

ctms/models.py
bsieber-mozilla, line 1: (L#285-318) 

Do we want the Cascade to delete the Stripe entity information associated with the email_id being deleted? Or mark those rows in some fashion?

PR #551, opened 2023-02-16
Ref #548: make basket_token field optional in contact responses body

ahoneiser:
Follow-up of #548


leplatrem (commented):
Could you please be a bit more specific in the PR title? 

For example, _make `basket_token` field optional in contact responses body_, or something like that.

Those PR titles show up directly in our CHANGELOG, and need to help readers understand what changed concretely in the release.

I'm OK with the change. Although I would feel more comfortable if it had an accompanying test (eg. unset the contact basket token field and delete it).

PR #554, opened 2023-02-17
Start to use factories to generate test data

grahamalama:
We use [static data](https://github.com/mozilla-it/ctms-api/blob/main/tests/unit/sample_data.py) as test data throughout our tests. Some observations I've made about this data and the tests that use them:
- it makes test setup not obvious. I ask myself, "when a given test starts, what data is already present in the database?"
- they promote the use of ["magic numbers"](https://github.com/mozilla-it/ctms-api/search?q=%2267e52c77-950f-4f28-accb-bb3ea1a2c51a%22)
- they're used in [complex fixtures](https://github.com/mozilla-it/ctms-api/blob/23bf39231ab680a27f2518cee4f47186602fa28b/tests/unit/conftest.py#L244)
- The division between "minimal" and "maximal" contact seems arbitrary

In this PR, we would introduce the use of test factories with [Factory Boy](https://factoryboy.readthedocs.io/en/stable/index.html) and [pytest-factoryboy](https://pytest-factoryboy.readthedocs.io/en/latest/). I've done minimal setup in this PR to demonstrate a proof of concept. This involved adding a `scoped_session` for database tests as well as making two example factories and using them in two example tests.

If we like this pattern, we can use it going forward with the goal of making existing tests easier to understand and new tests easier to write.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -12,8 +12,8 @@ from alembic import config as alembic_config
 from fastapi.testclient import TestClient
 from prometheus_client import CollectorRegistry
 from pydantic import PostgresDsn
+from pytest_factoryboy import register
 from sqlalchemy import create_engine, event
-from sqlalchemy.orm import Session
 from sqlalchemy_utils.functions import create_database, database_exists, drop_database
 
 from ctms.app import app, get_api_client, get_db, get_metrics
@@ -37,6 +37,7 @@ from ctms.crud import (
     get_stripe_products,
     get_waitlists_by_email_id,
 )
+from ctms.database import ScopedSessionLocal, SessionLocal
 from ctms.schemas import (
     ApiClientSchema,
     ContactSchema,
@@ -50,6 +51,8 @@ from tests.unit.sample_data import (
     SAMPLE_MOST_MINIMAL,
     SAMPLE_STRIPE_DATA,
 )
+
+from . import factories
 
 MY_FOLDER = os.path.dirname(__file__)
 TEST_FOLDER = os.path.dirname(MY_FOLDER)
@@ -100,10 +103,11 @@ def engine(pytestconfig):
         drop_database(test_db_url)
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def connection(engine):
     """Return a connection to the database that rolls back automatically."""
     conn = engine.connect()
+    SessionLocal.configure(bind=conn)
     yield conn
     conn.close()
 
@@ -115,16 +119,17 @@ def dbsession(connection):
     Adapted from https://docs.sqlalchemy.org/en/14/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites
     """
     transaction = connection.begin()
-    session = Session(autocommit=False, autoflush=False, bind=connection)
-    nested = connection.begin_nested()
+    session = ScopedSessionLocal()
+    nested = session.begin_nested()
 
     # If the application code calls session.commit, it will end the nested
     # transaction. Need to start a new one when that happens.
-    @event.listens_for(session, "after_transaction_end")
-    def end_savepoint(session, transaction):
+    @event.listens_for(ScopedSessionLocal, "after_transaction_end")
+    def end_savepoint(*args):
         nonlocal nested
-        if not nested.is_active:
-            nested = connection.begin_nested()
+        if nested.is_active:
+            session.expire_all()
+            nested = session.begin_nested()
 
     yield session
     session.close()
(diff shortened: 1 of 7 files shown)


tests/unit/conftest.py
```diff
@@ -197,6 +200,15 @@ def acoustic_newsletters_mapping(dbsession):
     return {r.source: r.destination for r in records}
 
 
+@pytest.fixture
+def test_factories(dbsession):
+    for factory in factories.__all__:
+        setattr(
+            attrgetter(f"{factory}._meta")(factories), "sqlalchemy_session", dbsession
+        )
+    return factories
```
grahamalama, line 209: As referenced in the PR description, this is how I ended up adding the `dbsession` fixture to the factories. I don't love it.

leplatrem (approved):
I like this! 

Since migrating the static data to this will require some efforts, what are the next steps? Merge this and live with the 2 ways, and migrate later?

grahamalama:
> Merge this and live with the 2 ways, and migrate later?

Potentially. I don't know if this is a viable solution, but we could make it a point in PR review to say, "Did you modify a test"? If so, replace any static fixture data with a factory."

tests/unit/conftest.py
```diff
@@ -115,16 +119,17 @@ def dbsession(connection):
...
     # If the application code calls session.commit, it will end the nested
     # transaction. Need to start a new one when that happens.
-    @event.listens_for(session, "after_transaction_end")
-    def end_savepoint(session, transaction):
+    @event.listens_for(ScopedSessionLocal, "after_transaction_end")
+    def end_savepoint(*args):
         nonlocal nested
-        if not nested.is_active:
-            nested = connection.begin_nested()
+        if nested.is_active:
```
grahamalama, line 130: @leplatrem I'm looking at this again, and frankly I'm confused why the boolean flip here is still working.
From:
```
if not nested.is_active:
```
to
```
if nested.is_active:
```

This works, and makes more sense:
```python
    nested = session.begin_nested()

    @event.listens_for(ScopedSessionLocal, "after_transaction_end")
    def end_savepoint(_, trx):
        nonlocal nested
        if trx.nested and not nested.is_active:
            session.expire_all()
            nested = session.begin_nested()
```

But I'm not entirely sure what's happening here to make this work or not work.

grahamalama: Ahh, I'm able to use the pattern straight from the docs
```python
def dbsession(connection):
    """Return a database session that rolls back.

    Adapted from https://docs.sqlalchemy.org/en/14/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites
    """
    transaction = connection.begin()
    session = ScopedSessionLocal()
    nested = connection.begin_nested()

    # If the application code calls session.commit, it will end the nested
    # transaction. Need to start a new one when that happens.
    @event.listens_for(ScopedSessionLocal, "after_transaction_end")
    def end_savepoint(*args):
        nonlocal nested
        if not nested.is_active:
            nested = connection.begin_nested()

    yield session
    session.close()
    transaction.rollback()
```

If I comment out [this line ](https://github.com/mozilla-it/ctms-api/blob/69750e77d3a7393af2721755e5187c3418555619/tests/unit/test_crud.py#L51)

tests/unit/factories.py
```diff
@@ -0,0 +1,54 @@
...
+    email_id = factory.Faker("uuid4")
+    primary_email = factory.Faker("email")
+    basket_token = factory.Faker("uuid4")
+    first_name = factory.Faker("first_name")
+    last_name = factory.Faker("last_name")
+    mailing_country = factory.Faker("country_code")
+    email_format = "T"
+    email_lang = factory.Faker("language_code")
+    double_opt_in = False
+    has_opted_out_of_email = False
```
bsieber-mozilla, line 40: Do we need to modify this for active/inactive contacts? _(I think active contacts should have `double_opt_in` to True)_

tests/unit/test_api_get.py
```diff
@@ -57,37 +60,13 @@ def test_get_ctms_for_minimal_contact(client, minimal_contact):
...
-                "subscribed": True,
-                "unsub_reason": None,
-            },
+                "format": newsletter.format,
+                "lang": newsletter.lang,
+                "name": newsletter.name,
+                "source": newsletter.source,
+                "subscribed": newsletter.subscribed,
+                "unsub_reason": newsletter.unsub_reason,
+            }
```
bsieber-mozilla, line 69: 👏

tests/unit/factories.py
grahamalama: I believe we will have to override this default for different test cases -- either we keep this as is and override `double_opt_in` where applicable, or vice versa. 

We could also create additional factories with different defaults for different test cases.

PR #566, opened 2023-03-03
Do not enforce Python patch version

leplatrem:
Can't run the project with 3.11.2 otherwise


bsieber-mozilla (approved):
👍 LGTM

_(Looks like a newline-change at the end of the pyproject.toml--formatter?)_

PR #567, opened 2023-03-03
Force UTC at SQLAlchemy client/session level

leplatrem:
My tests were failing when running locally.

I tried several things at the DB level ([1](https://stackoverflow.com/a/51697579/141895), [2](https://github.com/Kinto/kinto/blob/af30f2d066d5b007bfac15510843b93e66e7c8cf/docs/community.rst#L127)), and apparently setting `TZ=UTC` in `bin/test.sh` for the client was not enough. 

This patch solves it for me, since it forces the timezone at connection... 🤷

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -86,7 +86,11 @@ def engine(pytestconfig):
         create_database(test_db_url)
 
     echo = Settings().log_sqlalchemy or pytestconfig.getoption("verbose") > 2
-    test_engine = create_engine(test_db_url, echo=echo)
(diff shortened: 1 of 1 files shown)


grahamalama (approved):
Very strange. I don't like that I don't know why `export TZ=UTC` in `bin/test.sh` was enough for me, and not for you. I'm glad this fixes it, but I'm just wondering if there's some "root cause" that's worth investigating.

Also, this also makes the case for avoiding hardcoded timestamps in tests 😉.

Issue #565, opened 2023-03-01
CTMS-146: Add a command to detect or fix diffs between CTMS and Acoustic


leplatrem:
https://mozilla-hub.atlassian.net/browse/CTMS-146

If a user unsubscribes from a newsletter but for any reason the Acoustic data is not in sync, they will continue to receive emails.

We need a consistency-check command, that should report and ideally fix the diffs between CTMS and Acoustic

grahamalama:
Maybe something for another issue, but perhaps we could also leverage [Acoustic Webhooks](https://help.goacoustic.com/hc/en-us/articles/360039881833-Create-custom-workflows-with-webhooks) here to keep things in sync? So that if a change is made on the Acoustic side, it's propagated to CTMS.

bsieber-mozilla:
There is some ongoing effort in acoustic to clean up the system with a contractor--they had a single session last week. The documents in this drive show some of the current effort: https://docs.google.com/spreadsheets/d/1rWMlkvLlbZhDXX4ZYvj4QJ7j0k1JYlXR/edit?usp=share_link&ouid=118205549745087972930&rtpof=true&sd=true

leplatrem:
This command is very likely to never exist. The dataset is too big in front of the limited features of the Acoustic API. 

We've decided to change the strategy:
- fail loudly on synchronization errors
- use `acoustic resync` (#588) 
- or use `acoustic.py dump` (#686) to dump big parts of the database and re-upload to Acoustic FTP

Let's mark this as closed. Feel free to re-open if good reasons to do so

PR #569, opened 2023-03-03
Ref #565: CTMS-146: Add a way to force Acoustic resync of a list of contacts

leplatrem:
Ref #565 

From the first investigations, it does not seem like the Acoustic APi provides a way to fetch contact emails from queries. Only count matches. 
As an intermediary (and cheap) step, I added options to the `acoustic_sync` command so that an operator can force a resync of a list of contacts.

* [x] First implementation
* [x] Documentation
* [x] Make sure the Acoustic sync service would effectively rewrites all data --> [yes for `sub_*` fields](https://github.com/mozilla-it/ctms-api/blob/055c3deb683923a161e3eb8c712b53a1a1378f61/ctms/acoustic_service.py#L221-L224) **but** [not clear whether relational table newsletter are removed or not](https://github.com/mozilla-it/ctms-api/blob/055c3deb683923a161e3eb8c712b53a1a1378f61/ctms/acoustic_service.py#L371-L380)

### DIFF
--- a/ctms/bin/acoustic.py
+++ b/ctms/bin/acoustic.py
@@ -20,7 +20,6 @@ from ctms.log import configure_logging
 @click.group()
 @click.pass_context
 def cli(ctx):
-    ctx.obj["dbsession"] = SessionLocal()
     settings = config.Settings()
     configure_logging(logging_level=settings.logging_level.name)
     init_sentry()
@@ -38,7 +37,8 @@ def resync(
     waitlist: Optional[str] = None,
 ):
     """CTMS command to sync contacts with Acoustic."""
-    return do_resync(ctx.obj["dbsession"], emails_file, newsletter, waitlist)
+    with SessionLocal() as dbsession:
+        return do_resync(dbsession, emails_file, newsletter, waitlist)
 
 
 def do_resync(dbsession, emails_file=None, newsletter=None, waitlist=None):
--- a/ctms/bin/acoustic.py
+++ b/ctms/bin/acoustic.py
@@ -15,8 +15,6 @@ from ctms.database import SessionLocal
 from ctms.exception_capture import init_sentry
 from ctms.log import configure_logging
 
-logger = structlog.get_logger("ctms.bin.acoustic_resync")
-
 
 @click.group()
 @click.pass_context
@@ -55,7 +53,7 @@ def do_resync(dbsession, email_list=None, newsletter=None, waitlist=None):
             raise ValueError(f"Unknown waitlist {waitlist!r}")
         to_resync.extend(c.email.primary_email for c in contacts)
 
-    logger.info("Force resync of %s contacts", len(to_resync))
+    print("Force resync of %s contacts", len(to_resync))
     bulk_schedule_acoustic_records(dbsession, to_resync)
     dbsession.commit()
     return 0
--- a/ctms/bin/acoustic.py
+++ b/ctms/bin/acoustic.py
@@ -17,6 +17,13 @@ from ctms.exception_capture import init_sentry
 from ctms.log import configure_logging
 
 
+def confirm(msg):
+    answer = None
+    while answer not in ("", "y", "n"):
+        answer = input(f"{msg} y/N ").lower()
+    return answer == "y"
+
+
 @click.group()
 @click.pass_context
 def cli(ctx):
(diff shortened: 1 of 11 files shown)


grahamalama:
> [yes for sub_* fields](https://github.com/mozilla-it/ctms-api/blob/055c3deb683923a161e3eb8c712b53a1a1378f61/ctms/acoustic_service.py#L221-L224) but [not clear whether relational table newsletter are removed or not](https://github.com/mozilla-it/ctms-api/blob/055c3deb683923a161e3eb8c712b53a1a1378f61/ctms/acoustic_service.py#L371-L380)

Yeah, it would seem that we'd need to call [DeleteRelationalTableData](https://developer.goacoustic.com/acoustic-campaign/reference/deleterelationaltabledata) somehow during this process, but we don't. It would appear that `pysilverpop` doesn't even have such a method.

bsieber-mozilla:
>     (Maybe this was intentional as an archive for newsletter activity over time?
> 
> Yes, I believe that is correct.  This RT gives newsletter activity over time, and MoFo doesn’t use it in any queries for sends.  i.e. a row in that RT doesn’t determine in any way whether you are actually sent an email from us.  
> 
> MoFo (and I believe MoCo) uses 3 main values to determine whether you are “mailable” - and all 3 are on the Main Contact Table
> 
>      has_opted_out_of_email is equal to 0
> 
>     double_opt_in is equal to 1
> 
>     sub_* is equal to 1 [subscribed to the given list]
> 
> In this incident, the Acoustic values for the Subscriber weren’t changed.  After un-subbing via Preference Center, those values in Acoustic should =
> 
>     has_opted_out_of_email is equal to 1
> 
>     double_opt_in is equal to 1
> 
>     sub_* is equal to 0 [subscribed to the given list]
> 
> There may be other fields that are altered on an usub that I’m not aware of, but those are the 3 that are primarily responsible for determining whether an address gets sent to.

> From: https://mozilla-hub.atlassian.net/browse/CTMS-146?focusedCommentId=655835

bsieber-mozilla:
Another part of this resync; how does this relate to contact's that are stuck in a pending_state with more than the allotted retries? (Or should that be handled elsewhere?)

https://earthangel-b40313e5.influxcloud.net/d/7FxRQoNVz/ctms-prod-dashboard?orgId=1&viewPanel=30&from=1677790733274&to=1678136333274

leplatrem:
> This RT gives newsletter activity over time, and MoFo doesn’t use it in any queries for sends

This clarifies the missing part for this pull request. And this PR is thus now ready to be reviewed.

> Another part of this resync; how does this relate to contact's that are stuck in a pending_state with more than the allotted retries? (Or should that be handled elsewhere?)

I wasn't aware of this. Seeing the amount of pending records in Grafana is terrifying. This should definitely be discussed in a specific issue.

docs/deployment_guide.md
```diff
@@ -74,6 +74,29 @@ In order to add, list, or remove certain mappings, use the `acoustic_newsletters
...
+python ctms/bin/acoustic_sync.py --newsletter common-voice
+```
+
+Or a waitlist:
+```
+python ctms/bin/acoustic_sync.py --waitlist relay
+```
+
+See `python ctms/bin/acoustic_sync.py --help` for usage details.
+
```
bsieber-mozilla, line 99: Should this be run on the active background-job pod (restart with the new cli params)? or run as a separate process--then killed since it runs "indefinitely"?

leplatrem: You're right, that was confusing. I moved it to a separate command.

ctms/bin/acoustic.py
```diff
@@ -0,0 +1,69 @@
...
+)
+from ctms.database import SessionLocal
+from ctms.exception_capture import init_sentry
+from ctms.log import configure_logging
+
+
+@click.group()
+@click.pass_context
+def cli(ctx):
+    ctx.obj["dbsession"] = SessionLocal()
```
grahamalama, line 23: I think `ctx.obj["dbsession"] = SessionLocal()` will open a session that we never close, since we manually close it in the [`get_db`](https://github.com/mozilla-it/ctms-api/blob/998124cb794850a8017aecbb03a95f85fa42f35e/ctms/app.py#L124-L129) dependency.

What we could do is `ctx.obj["SessionLocal"] = SessionLocal`

Then below:
```python
with ctx.obj["SessionLocal"]() as dbsession:
    return do_resync(dbsession, emails_file, newsletter, waitlist)
```

Or, since SessionLocal is a factory anyway, I think we could just use the global directly:
```python
with SessionLocal() as dbsession:
    return do_resync(dbsession, emails_file, newsletter, waitlist)
```

ctms/bin/acoustic.py
```diff
@@ -0,0 +1,69 @@
...
+        to_resync.extend(c.email.primary_email for c in contacts)
+
+    if waitlist:
+        contacts = get_contacts_from_waitlist(dbsession, waitlist)
+        if not contacts:
+            raise ValueError(f"Unknown waitlist {waitlist!r}")
+        to_resync.extend(c.email.primary_email for c in contacts)
+
+    print("Force resync of %s contacts", len(to_resync))
+    bulk_schedule_acoustic_records(dbsession, to_resync)
```
grahamalama, line 63: Should we include a dry run mode and/or a confirmation prompt to the user? There could be some number (I don't know what that number is) where an operator might decide it's better to upload a CSV of updates to Acoustic directly instead of scheduling a bunch of contacts to be synced and waiting for the queue to drain out.

grahamalama (commented):
Looking good! Just a few more comments.

ctms/bin/acoustic.py
leplatrem: I realized that and then thought "ah it's a short-lived script anyway" 

But you're right, it's cleaner

ctms/bin/acoustic.py
leplatrem: Good idea

PR #572, opened 2023-03-06
 Ref #565: delete relational rows in Acoustic before inserting new ones 

leplatrem:
Ref #565

Acoustic should contain more data than CTMS does.

The relational table data should not be a "history trail".

Ideally, it should not contain more data than what is required for marketing to craft audience queries.
For now, start by adding a step that deletes from the relational table before inserting rows.

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -396,16 +396,24 @@ class CTMSToAcousticService:
                 self.context["main_status"] = status
                 self.context["main_duration_s"] = duration_s
 
-    def _insert_update_relational_table(self, table_name, rows):
-        if not rows:
-            return
+    def _replace_relational_table_data(self, table_name, email_id, rows):
         start_time = time.monotonic()
-        status = "success"
         table_id = self.relational_tables[table_name]
+        status = "success"
+        method = "delete_relational_table_data"
         try:
-            self.acoustic.insert_update_relational_table(
-                table_id=table_id, rows=rows
-            )  # Call to Acoustic
+            # Calls to Acoustic
+            self.acoustic.delete_relational_table_data(
+                table_id=table_id,
+                rows=[
+                    {"email_id": email_id},
+                ],
+            )
+            if rows:
+                method = "insert_update_relational_table"
+                self.acoustic.insert_update_relational_table(
+                    table_id=table_id, rows=rows
+                )
         except Exception:  # pylint: disable=broad-except
             status = "failure"
             raise
@@ -414,7 +422,7 @@ class CTMSToAcousticService:
                 duration = time.monotonic() - start_time
                 duration_s = round(duration, 3)
                 metric_params = {
-                    "method": "insert_update_relational_table",
+                    "method": method,
                     "status": status,
                     "table": table_name,
                 }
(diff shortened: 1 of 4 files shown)


ctms/acoustic_service.py
```diff
@@ -396,16 +396,24 @@ def _add_contact(self, list_id=None, sync_fields=None, columns=None):
...
+                table_id=table_id,
+                rows=[
+                    {"email_id": email_id},
+                ],
+            )
+            if rows:
+                method = "insert_update_relational_table"
+                self.acoustic.insert_update_relational_table(
+                    table_id=table_id, rows=rows
+                )
```
bsieber-mozilla, line 416: So we delete all `newsletter_rt` data associated with `email_id`; then 
if newsletter data exists _("existing" newsletter data stored in CTMS)_ for `email_id` insert all data into `newsletter_rt`?

It looks like we'd be updating the timestamps (_`create/update`_) in the [newsletter_converter](https://github.com/mozilla-it/ctms-api/blob/11610ffca765f320d37412cceffbfef18795fba7/ctms/acoustic_service.py#L217) 
https://github.com/mozilla-it/ctms-api/blob/11610ffca765f320d37412cceffbfef18795fba7/ctms/acoustic_service.py#L240-L246

bsieber-mozilla: ~It looks like we'd be updating the timestamps (_`create/update`_) in the [newsletter_converter](https://github.com/mozilla-it/ctms-api/blob/11610ffca765f320d37412cceffbfef18795fba7/ctms/acoustic_service.py#L217) 
https://github.com/mozilla-it/ctms-api/blob/11610ffca765f320d37412cceffbfef18795fba7/ctms/acoustic_service.py#L240-L246~

_Misread the dict.get call. Ignore this!_

pyproject.toml
```diff
@@ -25,7 +25,7 @@ passlib = {extras = ["argon2"], version = "^1.7.4"}
 python-dateutil = "^2.8.1"
 dockerflow = "^2022.8.0"
 sentry-sdk = "^1.14.0"
-pysilverpop = "^0.2.6"
+pysilverpop = {git = "https://github.com/leplatrem/pysilverpop.git", rev = "add-delete-relational-table-data"}
```
bsieber-mozilla, line 28: https://github.com/theatlantic/pysilverpop/pull/6/files
👍 Thanks @leplatrem

bsieber-mozilla (approved):

ctms/acoustic_service.py
grahamalama, line 416: What happens if something goes wrong between deleting all rows and re-adding them? Will we attempt to retry?

Can we attempt to say, only delete the RT rows for unsubscribed newsletters? I think we might have to do this anyway, since [the docs say](https://developer.goacoustic.com/acoustic-campaign/reference/deleterelationaltabledata):

> If the table has multiple key columns, each ROW has multiple COLUMN elements. All key columns must be provided for each ROW.

Looking at the Acoustic newsletters RT schema, both `email_id` and `newsletter_name` have the little key icon in the UID column
image

So unless you tried this and it worked already, we might have to provide the newsletter name for each row we want to delete anyway.

leplatrem: You're right! 

This PR was a rabbit hole

PR #583, opened 2023-03-08
Fix #582: force resync of newsletter contacts when mapping is added

### LINKED ISSUE #582: Resync all newsletters contacts when a new mapping column is added
When we add a new Acoustic column and add a new mapping for a newsletter ( #456 ), all contacts that were subscribed to the newsletter need to be resynced.

leplatrem:
Fix #582

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -10,6 +10,7 @@ from pydantic import UUID4
 from sqlalchemy import asc, or_
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session, joinedload, load_only, selectinload
+from sqlalchemy.sql.expression import true
 
 from .auth import hash_password
 from .backport_legacy_waitlists import format_legacy_vpn_relay_waitlist_input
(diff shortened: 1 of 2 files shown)


ctms/crud.py
```diff
@@ -932,6 +933,17 @@ def create_acoustic_newsletters_mapping(dbsession, source, destination):
     row = AcousticNewsletterMapping(source=source, destination=destination)
     # This will fail if the mapping already exists.
     dbsession.add(row)
+
+    # Force a resync of all contacts that are subscribed to this newsletter.
```
grahamalama, line 937: It feels as though this should be in a separate CRUD function, (maybe `sync_newsletter_with_acoustic`) since the management of newsletter mappings and when they're synced seems like separate concerns.

But, it does make sense to attempt to sync a newsletter just after it's added. 

The dream code version of a management CLI (to me) seems like it'd be:
```
add <newsletter name>:<sub_ column>
add <newsletter name>:<sub_ column> --sync
sync <newsletter name>
```

leplatrem: Yes, I agree.

Let's ditch this command and add an `acoustic` CLI with subcommands in #569

PR #587, opened 2023-03-09
Fix #581: polish sync logging and store last error in DB

### LINKED ISSUE #581: Log Acoustic sync errors for better troubleshooting
Ref #575 
Ref #569 

Currently, we count the number of retries when a record fails to be sync to Acoustic, but we don't log/save the error. This makes troubleshooting impossible.

leplatrem:
Fix #581

This should put us in a slightly better place regarding invalid email addresses. We also don't silent random errors during sync, we only catch Acoustic errors.


ctms/acoustic_service.py
```diff
@@ -459,12 +463,14 @@ def attempt_to_upload_ctms_contact(
...
-        except (SilverpopResponseException, Timeout):
+        except (SilverpopResponseException, Timeout) as exc:
             # failure
             self.logger.exception(
                 "Failure for contact in sync to acoustic...",
                 success=False,
                 **self.context,
             )
-            return False
+            # Silverpop is supposed to provide the fault string as message.
```
bsieber-mozilla, line 473: thanks for the notes here!

bsieber-mozilla (approved):

PR #588, opened 2023-03-09
Ref #565: Reset `retry` status from `acoustic.py resync` command

leplatrem:
This comes on top of #569, only last commit should be reviewed :)


bsieber-mozilla (approved):
👍

PR #627, opened 2023-04-10
Use factory to create `maximal_contact`, use in test

grahamalama:
This PR:
- Adds factories for AMO, MoFo, and FXA models
- Incorporates these into the Email factory
- Uses this new factory in a test

I realize in this PR that the maximal_contact fixtures "shadows" the sample data. This will hopefully be remedied soon -- the ultimate goal will be to create a `maximal_contact` (any really, any contact) like so:

```python
email_factory(
    amo=True,
    fxa=True,
    mofo=True,
    newsletters=2,
    waitlists=2,
)
```

and use that in tests.


grahamalama:
Closing this for now -- I think the first thing we should do is move everything to fixtures (and away from static data in `sample_data`), then we can use factories

PR #628, opened 2023-04-11
Replace most sample data with pytest fixtures

grahamalama:
This PR moves most contact fixture data from dicts + the `ContactVendor` class in `tests/unit/sample_data.py` to `pytest` fixtures in `conftest.py`

For each contact sample, there are two fixtures:
- with a `_data` suffix that just creates the schema object
- without a `_data` suffix, which persists the data to the database

this allows for compatibility with preexisting fixtures.

The follow-up to this PR will be migrating the `FAKE` / `SAMPLE` Stripe data to `conftest` and fully deleting `sample_data.py`. Then, we can turn our attention to how we use the fixture data in tests. Where it's appropriate, we can use factories instead of static fixtures directly in test setup to make them more obvious.


tests/unit/test_api_put.py
```diff
@@ -5,9 +5,13 @@
...
 from structlog.testing import capture_logs
 
-from tests.unit.sample_data import SAMPLE_CONTACTS
+from tests.unit.conftest import SAMPLE_CONTACT_PARAMS
 from tests.unit.test_api import _compare_written_contacts
 
+POST_TEST_CASES = pytest.mark.parametrize(
+    "put_contact", SAMPLE_CONTACT_PARAMS, indirect=True
+)
+
```
bsieber-mozilla, line 14:
```suggestion
PUT_TEST_CASES = pytest.mark.parametrize(
    "put_contact", SAMPLE_CONTACT_PARAMS, indirect=True
)

```

tests/unit/sample_data.py
```diff
@@ -245,7 +26,8 @@ def fake_stripe_id(prefix: str, seed: str, suffix: Optional[str] = None) -> str:
     "Customer": {
         "stripe_id": FAKE_STRIPE_ID["Customer"],
         "stripe_created": datetime(2021, 10, 25, 15, 34, tzinfo=timezone.utc),
-        "fxa_id": SAMPLE_EXAMPLE.fxa.fxa_id,
+        # TODO magic string from fxa schema
+        "fxa_id": "6eb6ed6ac3b64259968aa490c6c0b9df",
```
bsieber-mozilla, line 30: _TODO here_

bsieber-mozilla (approved):
Nice! Thanks for this PR

tests/unit/test_api_put.py
grahamalama: Good catch 👍🏻 [7bfc447](https://github.com/mozilla-it/ctms-api/pull/628/commits/7bfc4477cfe32f5e1066086e3abc1f809df69304)

PR #629, opened 2023-04-12
Migrate `SAMPLE_STRIPE_DATA` to pytest fixtures

grahamalama:
A follow-on of #628, in this PR, we move more data from `sample_data.py` to pytest fixtures in `conftest.py`, and use those fixtures in tests.

After this PR, there will be another that somehow moves `FAKE_STRIPE_ID` / `fake_stripe_id()` out of sample_data.py, then we can delete that module

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,7 +1,9 @@
 """pytest fixtures for the CTMS app"""
 import json
 import os.path
+from datetime import datetime, timezone
 from glob import glob
+from time import mktime
 from typing import Callable, Optional
 from unittest import mock
 from uuid import UUID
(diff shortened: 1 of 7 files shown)


leplatrem (approved):
💯

bsieber-mozilla (approved):
LGTM; thanks for cleaning up the testing strategy

tests/unit/conftest.py
```diff
@@ -673,26 +681,235 @@ def stripe_test_json(request):
 
 
 @pytest.fixture
-def contact_with_stripe_customer(dbsession, example_contact):
+def stripe_customer_data():
+    return {
+        "stripe_id": FAKE_STRIPE_ID["Customer"],
+        "stripe_created": datetime(2021, 10, 25, 15, 34, tzinfo=timezone.utc),
+        # TODO magic string from fxa schema
+        "fxa_id": "6eb6ed6ac3b64259968aa490c6c0b9df",
```
bsieber-mozilla, line 689: _TODO_ here

bsieber-mozilla: Resolving with this https://github.com/mozilla-it/ctms-api/pull/628#discussion_r1164059629 in mind

PR #632, opened 2023-04-13
Finish moving Stripe data out of `sample_data.py`

grahamalama:
Follow up of #628 and #629 

The overall goal of this PR was to move all of the remaining functions / data out of `sample_data.py` to delete that module. Here's how we did that:

- Use fixture data (instead of `FAKE_STRIPE_ID`) to make assertions. 
  - In some cases, we had to pull in fixtures that weren't already provided. This will hopefully be a temporary measure while we refactor tightly coupled fixtures 
- Convert FAKE_STRIPE_ID to constants in `conftest.py`
- Use `fake_stripe_id` in tests and fixtures that don't need `FAKE_STRIPE_ID`


leplatrem (approved):
LGTM (GG on removing sample_data.py 🎉)

Issue #633, opened 2023-04-14
Replace static fixtures with factory-generated fixtures


grahamalama:
This issue will track the work of migrating test objects from interdependent pytest fixtures and static data to factory generated fixtures.

PR #634, opened 2023-04-14
Create Stripe customer factories, use in tests

grahamalama:
Ref: #633 

Deletes these fixtures:
- `stripe_customer_data`
- `raw_stripe_customer_data`
- `contact_with_stripe_customer`

In favor of factory usage in tests

### DIFF
--- a/tests/factories/models.py
+++ b/tests/factories/models.py
@@ -47,8 +47,8 @@ class FirefoxAccountFactory(BaseSQLAlchemyModelFactory):
     primary_email = factory.SelfAttribute("email.primary_email")
     created_date = factory.Faker("date")
     lang = factory.Faker("language_code")
-    first_service = factory.Faker("random_element", elements=["Firefox"])
-    account_deleted = factory.Faker("boolean", chance_of_getting_true=30)
+    first_service = None
+    account_deleted = False
 
     email = factory.SubFactory(factory="tests.factories.models.EmailFactory")
 
--- a/tests/factories/models.py
+++ b/tests/factories/models.py
@@ -1,8 +1,12 @@
+from datetime import datetime, timezone
+from uuid import uuid4
+
 import factory
 from factory.alchemy import SQLAlchemyModelFactory
 
 from ctms import models
 from ctms.database import ScopedSessionLocal
+from tests.data import fake_stripe_id
 
 
 class BaseSQLAlchemyModelFactory(SQLAlchemyModelFactory):
@@ -35,6 +39,20 @@ class WaitlistFactory(BaseSQLAlchemyModelFactory):
     email = factory.SubFactory(factory="tests.factories.models.EmailFactory")
 
 
+class FirefoxAccountFactory(BaseSQLAlchemyModelFactory):
+    class Meta:
+        model = models.FirefoxAccount
+
+    fxa_id = factory.LazyFunction(lambda: uuid4().hex)
+    primary_email = factory.SelfAttribute("email.primary_email")
+    created_date = factory.Faker("date")
+    lang = factory.Faker("language_code")
+    first_service = factory.Faker("random_element", elements=["Firefox"])
+    account_deleted = factory.Faker("boolean", chance_of_getting_true=30)
+
+    email = factory.SubFactory(factory="tests.factories.models.EmailFactory")
+
+
 class EmailFactory(BaseSQLAlchemyModelFactory):
     class Meta:
         model = models.Email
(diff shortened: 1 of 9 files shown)


bsieber-mozilla (commented):
```
The job running on runner GitHub Actions 16 has exceeded the maximum execution time of 5 minutes.
```

Seems like the tests might've ran a little longer than normal? (Or maybe this was a CI-hiccup?)

---Does this run long on local as well?

tests/factories/models.py
```diff
@@ -58,9 +76,36 @@ def newsletters(self, create, extracted, **kwargs):
             for _ in range(extracted):
                 NewsletterFactory(email=self, **kwargs)
 
+    @factory.post_generation
+    def fxa(self, create, extracted, **kwargs):
+        if not create:
+            return
+        if extracted:
+            FirefoxAccountFactory(email=self, **kwargs)
```
leplatrem, line 84: missing return?

tests/factories/models.py
```diff
@@ -35,6 +39,20 @@ class Meta:
     email = factory.SubFactory(factory="tests.factories.models.EmailFactory")
 
 
+class FirefoxAccountFactory(BaseSQLAlchemyModelFactory):
+    class Meta:
+        model = models.FirefoxAccount
+
+    fxa_id = factory.LazyFunction(lambda: uuid4().hex)
+    primary_email = factory.SelfAttribute("email.primary_email")
```
leplatrem, line 47: 👌 
Shall we add a comment that this simulates the DB relationship?

tests/factories/models.py
```diff
@@ -35,6 +39,20 @@ class Meta:
...
+    created_date = factory.Faker("date")
+    lang = factory.Faker("language_code")
+    first_service = factory.Faker("random_element", elements=["Firefox"])
+    account_deleted = factory.Faker("boolean", chance_of_getting_true=30)
```
leplatrem, line 51: What's the value of not setting it always to False?

tests/unit/test_ingest_stripe.py
```diff
@@ -653,9 +650,10 @@ def test_get_email_id_subscription(
 
 
 def test_get_email_id_invoice(
-    dbsession, contact_with_stripe_customer, raw_stripe_invoice_data
+    dbsession, stripe_customer_factory, raw_stripe_invoice_data
 ):
     """A Stripe Invoice and related objects can return the related email_id."""
+    customer = stripe_customer_factory(stripe_id=FAKE_STRIPE_CUSTOMER_ID)
```
leplatrem, line 656: Why do we set the id here?

leplatrem (approved):

tests/unit/test_ingest_stripe.py
grahamalama: While we're in the middle of conversion, there are still a few places where we have "magic" object IDs. In this case, the way we use `raw_stripe_invoice_data` assumes that there's a a customer whose ID is `FAKE_STRIPE_CUSTOMER_ID`. So for now, that's what we specify when we generate the Stripe customer, but in a follow-up PR, we'll hopefully be able to remove this.

tests/factories/models.py
grahamalama: There is none -- bad copypasta 🤦🏻‍♂️. [34ac959](https://github.com/mozilla-it/ctms-api/pull/634/commits/34ac9595c94eab3a1639d39562e776106917e26a)

tests/factories/models.py
grahamalama: Nope -- since this is the `EmailFactory`, all we're doing here is creating the `FirefoxAccount` and associating it with this Email object. In a test, this would work fine:

```
email = EmailFactory()
dbsession.commit()
assert email.fxa ...
```

Unless you're saying that we should avoid the implicit return of `None`, in which case I could make that fix.

leplatrem: I get it, `email=self` does it. Sorry for the confusion

tests/factories/models.py
leplatrem: I'm confused by my own comment...

I didn't know about `SelfAttribute` and looked up its docs to understand how it was leveraged. I first understood that we were linking the two fields using this lazy attribute, as if they were linked by foreign keys in the DB. 

But actually, I see that `primary_email` in the `FirefoxAccount` model is not linked to `email.primary_email` at all, they can be different AFAIU. So scratch that

PR #640, opened 2023-04-19
Basket Rotation Script/Process

bsieber-mozilla:
The current/prior basket processes of rotation current involved multiple coordinated parties, with data export/imports. 

https://docs.google.com/document/d/1T7vAHrUSTU0yiy3YcxDqh5pOJ2omNza6oX8ViGwy7AA/edit#

----

Intent of PR: 
- Simplify coordination
- Reduce data export necessity (PII)
- No downtime of API (although we could for better db perf)

Questions:
- Is this unacceptable?

### DIFF
--- a/ctms/background_metrics.py
+++ b/ctms/background_metrics.py
@@ -112,6 +112,17 @@ class BackgroundMetricService:  # pylint: disable=too-many-instance-attributes
             registry=registry,
         )
 
+        self.token_rotation_gauge = Gauge(
+            name=metric_prefix + "token_rotation_gauge",
+            documentation="Gauge of the number of contacts to be updated.",
+            labelnames=[
+                "app_kubernetes_io_component",
+                "app_kubernetes_io_instance",
+                "app_kubernetes_io_name",
+            ],
+            registry=registry,
+        )
+
     def inc_acoustic_request_total(self, method, status, table):
         self.requests.labels(
             method=method,
@@ -176,3 +187,10 @@ class BackgroundMetricService:  # pylint: disable=too-many-instance-attributes
             app_kubernetes_io_instance=self.app_kubernetes_io_instance,
             app_kubernetes_io_name=self.app_kubernetes_io_name,
         ).set(duration_seconds)
+
+    def gauge_token_rotation(self, value):
+        self.token_rotation_gauge.labels(
+            app_kubernetes_io_component=self.app_kubernetes_io_component,
+            app_kubernetes_io_instance=self.app_kubernetes_io_instance,
+            app_kubernetes_io_name=self.app_kubernetes_io_name,
+        ).set(value)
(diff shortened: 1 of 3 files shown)


bsieber-mozilla:
I would love a basic review of this PR's intent. I think we might have most of the resources we need within CTMS--and not have to touch Acoustic (by using our background sync).

I pushed this commit with `no-verify` so theres some fixing for linters. 
(Also no tests currently)

leplatrem:
What would be the overall strategy?

Questions that pop up in my head?
- How do you distinguish the contacts that were already rotated?
- Looking at the maximum rate of contacts per seconds handled by the background job, and the total number of contacts, what would be an optimistic estimation of the time required to rotate them all?
- How can we make sure this massive resync won't interfere with regular newsletters/waitlists (un)subscriptions?

bsieber-mozilla:
> What would be the overall strategy?

The approach is to reduce PII data having to be exported, since we have the Acoustic sync job. 

1. (if required)  Similar to prior process -- Scale down/off service
2. Run script in CTMS
3. Observe metrics gauge and `failed_records` to determine records that requires retry.

Data exchange continue to happen through services, we can use the existing acoustic job. 

If we determine doing all email_ids at once is too costly, we could add some delay between loops--but that could mean all records to update takes a long time.

> Questions that pop up in my head?
> 
>     How do you distinguish the contacts that were already rotated?

We query for all contacts with non-null basket_tokens; then per-record log the success/failure and provide a metric gauge. The failed records are presented at the end of the script; success records could also be provided as output.

>     Looking at the maximum rate of contacts per seconds handled by the background job, and the total number of contacts, what would be an optimistic estimation of the time required to rotate them all?

Does this time estimation mean we are using all the DB resources and shut down the service, or we're ok with handling requests and having this job run over a few days possibly. 

>     How can we make sure this massive resync won't interfere with regular newsletters/waitlists (un)subscriptions?

We could do the same as what we did previously, shut down the service and call out the maintenance period.

---

What this script hopes to accomplish:
- reduce/remove PII on personal computers
- reduce/remove touches in DB (now scripted)
- use existing acoustic processing, instead of manual import job
- reduce coordination; limited SRE work request for script run, watch the gauge/logs

bsieber-mozilla:
On March 7th, it shows that CTMS was able to handle a lot of traffic. This metric might be false: 

Contacts sync'd to acoustic per min: max 3,455,407 contacts/min
Screenshot 2023-04-19 at 9 25 56 AM

bsieber-mozilla:
Looking into that last 40 days; a more reasonable max is 422.
8mil/422 = 18957.35 min = 316 hours = 13 days

grahamalama:
For your estimates @bsieber-mozilla, where is the estimate of `8mil` coming from?

bsieber-mozilla:
However, those times still seem to be unreasonable numbers; if we wanted to remove the schedule_acoustic_record call--this script can be converted to just the CTMS queries with split acoustic files as output. Or we use the  Acoustic SFTP apis?

bsieber-mozilla:
I'm going to hold on this PR--and just inquire for SRE support-> this is the checklist I'll be following: 
https://docs.google.com/document/d/1b4-PCGzeIUQJADMB286W29J0Im8cz5Z25nNUv83Q1Kw/edit#heading=h.v73t256gwyb3

PR #644, opened 2023-04-20
Adjust integration tests to Basket changes, fix bugs

grahamalama:
Adjusts some aspects of the integration tests so that they match the latest changes to basket.

There are still some issues with `to_vendor` (https://github.com/mozmeao/basket/issues/1012), so we'll need to add more test cases here when that's figured out so we can capture regressions.

### DIFF
--- a/tests/integration/test_basket_waitlist_subscription.py
+++ b/tests/integration/test_basket_waitlist_subscription.py
@@ -1,5 +1,4 @@
 import os
-import time
 from uuid import uuid4
 
 import backoff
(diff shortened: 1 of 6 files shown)


tests/integration/test_basket_waitlist_subscription.py
```diff
@@ -217,18 +229,30 @@ def check_subscribed():
...
+            "name": "relay-vpn-bundle",
+            "source": "https://relay.firefox.com/vpn-relay/waitlist/",
+            "fields": {
+                "geo": "fr",
+            },
+        },
+    ]
     assert contact_details["relay_waitlist"] == {
-        "geo": "fr",
+        "geo": "es",
```
leplatrem, line 250: Shall we add a comment here that it takes the last one of the list here? https://github.com/mozilla-it/ctms-api/blob/23bf39231ab680a27f2518cee4f47186602fa28b/ctms/schemas/contact.py#L235

leplatrem (approved):

tests/integration/test_basket_waitlist_subscription.py
grahamalama: Yes -- I think I was hesitant because I wasn't sure that it'd always be the last.

grahamalama: Separately, Should we also modify this behavior in the future to make it so that the country is taken from the waitlist that was most recently updated?

grahamalama: Wait actually, now I'm confused again. We expect that it'd be the last one of the list, but that assertion is saying that the country is `"es"`, when the "last one" is `"fr"`.

leplatrem: With the changes of this PR, it will take the first one. I'm sorry about the confusion

PR #654, opened 2023-04-27
Use validators for `CTMSResponse` intialization

grahamalama:
This PR adds `pydantic` `@validator` methods to provide defaults when initializing CTMSResponse models, instead of doing it inline.

It also uses a `root_validator` to initialize the legacy VPN waitlist fields. I don't necessarily prefer that method over using `__init__`, but it does feel like more of a standard pydantic pattern.


leplatrem (approved):
🤩

bsieber-mozilla (commented):
🥇 very nice

PR #661, opened 2023-05-03
Remove indirection in getting contacts

grahamalama:
This PR tweaks the way we query for contacts by `email_id` or my multiple IDs. Now, `get_contacts_by_email_id` returns a `Contact` object directly, and we combine the various functions we had to query multiple contacts into one `get_contacts_by_any_id` function.


leplatrem (approved):
Slick 🤩 

From the changed lines, I couldn't spot any potential query degradation 💯

PR #663, opened 2023-05-05
Add method to generate a `Contact` from an `Email`

grahamalama:
Adds a `ContactSchema.from_email` method so that we can consolidate that logic (that is, generating a `Contact` from an `Email` object) in one place.

### DIFF
--- a/ctms/schemas/contact.py
+++ b/ctms/schemas/contact.py
@@ -1,5 +1,6 @@
+from collections import defaultdict
 from datetime import datetime
-from typing import List, Literal, Optional, Set, Union
+from typing import TYPE_CHECKING, List, Literal, Optional, Set, Union, cast
 from uuid import UUID
 
 from pydantic import AnyUrl, BaseModel, Field, root_validator, validator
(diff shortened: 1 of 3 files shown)


ctms/schemas/contact.py
```diff
@@ -27,6 +28,94 @@
...
+                "currency": price.currency,
+                "amount": price.unit_amount,
+                "interval_count": price.recurring_interval_count,
+                "interval": price.recurring_interval,
+            }
+            by_product[price.stripe_product_id].append(product_data)
+    return by_product
+
+
+def _product_metadata(subscriptions_by_product):
```
leplatrem, line 70: Adding a comment may help here, at least explaining that we pick the latest product from the subscription dict values

leplatrem (approved):
The purpose/content of `_product_metadata()` wasn't immediate without comment, but everything looks good 💯

ctms/schemas/contact.py
grahamalama: @leplatrem ended up going in a pretty different direction for these functions, so retagging you for review.

PR #669, opened 2023-05-11
Add fixture factories for `StripeInvoice` and `StripeInvoiceLineItem`

grahamalama:
Adds factories for `StripeInvoice` and `StripeInvoiceLineItem`, and starts to use them in tests.

The next PR in this series will modify some of the tests in `test_ingest_stripe`.

Addresses #633

### DIFF
--- a/tests/unit/test_crud.py
+++ b/tests/unit/test_crud.py
@@ -16,12 +16,6 @@ from ctms.crud import (
     create_fxa,
     create_mofo,
     create_newsletter,
-    create_stripe_customer,
-    create_stripe_invoice,
-    create_stripe_invoice_line_item,
-    create_stripe_price,
-    create_stripe_subscription,
-    create_stripe_subscription_item,
     delete_acoustic_field,
     delete_acoustic_newsletters_mapping,
     delete_acoustic_record,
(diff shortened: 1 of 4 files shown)


tests/unit/test_crud.py
```diff
@@ -646,22 +633,9 @@ def test_get_stripe_customer_by_nonexistent_fxa_id(
...
-    stripe_subscription_item_data,
-    stripe_invoice_data,
-    stripe_invoice_line_item_data,
-):
-    """
-    Create two complete trees of Stripe objects.
-
-    We don't use ForeignKeys for Stripe relations, because the object may come
-    out of order for foreign key contraints. This helps check that the manually
+class TestStripeRelations:
```
leplatrem, line 636: I see that `self` isn't used anywhere. 
Did you create a class to "group" the tests? Shall we move them to `test_crud_stripe_releations` instead?

leplatrem (approved):
💯

tests/unit/test_crud.py
grahamalama: I did, both to group the tests and so that I could define a fixture that's autoused _only_ in the scope of these tests.

> Shall we move them to `test_crud_stripe_releations` instead

Do you mean a new module i.e. `test_crud_stripe_releations.py`?

PR #703, opened 2023-06-15
Migrate integration tests to MariaDB v11

leplatrem:
Our integration tests started to fail 6 days ago.

mariadb v11 was released and we were using `:latest`. Pinning to v10 fixes the tests.

The upgrade notes don't seem to mention anything relevant https://mariadb.com/kb/en/upgrading-from-mariadb-10-11-to-mariadb-11-0/
But here it is mentionned that official images don't have the `mysql` alias anymore...
https://mariadb.com/kb/en/mariadb-11-0-1-release-notes/#docker-official-images

### DIFF
--- a/bin/integration-test.sh
+++ b/bin/integration-test.sh
@@ -12,7 +12,7 @@ DOCKER_COMPOSE=${DOCKER_COMPOSE-"docker-compose"}
 export TZ=UTC
 
 # Create newsletters in basket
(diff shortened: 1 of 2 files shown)


grahamalama (approved):
Nice :shipit: 

> mariadb v11 was released and we were using :latest. Pinning to v10 fixes the tests.

So it seems that we opted to not pin the container and instead update the command we use, i.e. `mariadb` instead of `mysql`?

leplatrem:
> So it seems that we opted to not pin the container and instead update the command we use, i.e. mariadb instead of mysql?

yes :)

PR #710, opened 2023-06-20
Fix #611, Ref #708: fix Acoustic newsletters subscription timestamps

### LINKED ISSUE #611: Data loss during conversion from `Model` to `Schema`? (timestamps)
https://github.com/mozilla-it/ctms-api/blob/c3d3223216007cd987bc9868fa9189e69a41ef81/ctms/acoustic_service.py#L248-L250

Unless the `update_timestamp` field is null in the contact.newsletter, this field might never update with the current logic (to reuse the prior timestamp).

I believe the `update_timestamp` field should be updated when a contact is updated.

---
The [NewsletterSchema](https://github.com/mozilla-it/ctms-api/blob/c3d3223216007cd987bc9868fa9189e69a41ef81/ctms/acoustic_service.py#L224) doesn't have the timestamps (_[but do exist on the model)](https://github.com/mozilla-it/ctms-api/blob/c3d3223216007cd987bc9868fa9189e69a41ef81/ctms/models.py#L96-L117)_--so, I assume, the timestamps are currently always updated to be the same value of "_today"?

During the conversion from model.Email to schema.ContactSchema--are we losing the timestamp information as well?

`(some) Models have timestamps -> (some) Schemas appear to not have timestamps`

leplatrem:
Fix #611, Ref #708

In this pull-request, I fix a piece of code that never worked: the newsletter subscription timestamps sent to Acoustic were never read from the database.

Basically schemas do not read thetimestamps fields from the DB. This could be a possible fix, but my confidence in its elegance is kind of low.

This codebase has a really dark power of demotivation. Especially if you ever worked with tools like Django Rest Framework.

Plus, most test case assertions are not designed in a flexible way.

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -10,6 +10,9 @@ from sqlalchemy import asc, or_
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session, joinedload, load_only, selectinload
 
+from ctms.schemas.newsletter import NewsletterTableSchema
+from ctms.schemas.waitlist import WaitlistTableSchema
+
 from .auth import hash_password
 from .backport_legacy_waitlists import format_legacy_vpn_relay_waitlist_input
 from .database import Base
@@ -173,7 +176,22 @@ def get_contact_by_email_id(db: Session, email_id: UUID4) -> Optional[ContactSch
     email = get_email(db, email_id)
     if email is None:
         return None
-    return ContactSchema.from_email(email)
+    contact = ContactSchema.from_email(email)
+
+    # In the `ContactSchema` instance, the newsletters and waitlists are
+    # instances of `NewsletterSchema` and `WaitlistSchema`,
+    # which loose the `create_timestamp` and `update_timestamp` fields.
+    # Note: the whole code base with regards to models and schemas is a real house of cards.
+    contact_newsletters = get_newsletters_by_email_id(db, email_id)
+    contact.newsletters = [
+        NewsletterTableSchema.from_newsletter(nl) for nl in contact_newsletters
+    ]
+    contact_waitlists = get_waitlists_by_email_id(db, email_id)
+    contact.waitlists = [
+        WaitlistTableSchema.from_waitlist(wl) for wl in contact_waitlists
+    ]
+
+    return contact
 
 
 def get_contacts_by_any_id(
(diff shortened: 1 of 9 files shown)


ctms/crud.py
```diff
@@ -173,7 +176,22 @@ def get_contact_by_email_id(db: Session, email_id: UUID4) -> Optional[ContactSch
...
+    # which loose the `create_timestamp` and `update_timestamp` fields.
+    # Note: the whole code base with regards to models and schemas is a real house of cards.
+    contact_newsletters = get_newsletters_by_email_id(db, email_id)
+    contact.newsletters = [
+        NewsletterTableSchema.from_newsletter(nl) for nl in contact_newsletters
+    ]
+    contact_waitlists = get_waitlists_by_email_id(db, email_id)
+    contact.waitlists = [
+        WaitlistTableSchema.from_waitlist(wl) for wl in contact_waitlists
+    ]
```
grahamalama, line 192: I'm kind of putting my entire initial review here because I think this section represents the "main idea" of this PR -- when we build a contact, include `created` and `updated` timestamps with waitlists and newsletters. Here are my thoughts:

- Is there a reason we only want to do this waitlist and newsletter "patching" for contacts here in this function? It seems like we'd want to include these everywhere, so we may want to change the `ContactSchema` from

```python
class ContactSchema(ComparableBase):
    """A complete contact."""
    ...
    newsletters: List[NewsletterSchema] = []
    waitlists: List[WaitlistSchema] = []
```

to 

```python
class ContactSchema(ComparableBase):
    """A complete contact."""
    ...
    newsletters: List[NewsletterTableSchema] = []
    waitlists: List[WaitlistTableSchema] = []
```

or add the missing fields directly to the existing schemas. 

In fact, saying that this function returns a `ContactSchema` might be "cheating" a bit since we define a Contact's newsletters and waitlists as `NewsletterSchema`s and `WaitlistSchema`s respectively, but now we're returning different models.

- if that doesn't work and we only want these new schemas here, I think that if they are truly a 1:1 mapping to the SQLAlchemy models, we can use Pydantic's [ORM Mode](https://docs.pydantic.dev/latest/usage/models/#orm-mode-aka-arbitrary-class-instances) rather than creating these custom class methods to convert ORM objects to Pydantic models

leplatrem:
Replaced by #711

PR #711, opened 2023-06-26
Fix #611: `ContactSchema` now reflects what is stored in database

### LINKED ISSUE #611: Data loss during conversion from `Model` to `Schema`? (timestamps)

leplatrem:
This PR clarifies a bit that we are sometimes dealing with instances coming the database (`*Table*` schemas) and instances coming as input (`*In*` schemas).

It fixes bug #611 where timestamps were dropped during the Acoustinc sync process (because objects from the database were casted to input objects that don't have timestamp fields)


grahamalama (commented):
> instances coming the database (`*Table*` schemas)

I saw this and was slightly thrown by the naming introduced in d0e27b0256f0d6153b11a78296d924bbc22d7006, since there's no "Contact" table in the database. Is this to mean "a contact comprised entirely of `*TableSchema` objects?

ctms/crud.py
```diff
@@ -477,11 +477,12 @@ def create_waitlist(
...
         return None
-    if not isinstance(waitlist, WaitlistInSchema):
-        # Sample data are used as both input (`WaitlistInSchema`) and internal (`WaitlistSchema`)
-        # representations.
-        waitlist = WaitlistInSchema(**waitlist.dict())
-    db_waitlist = Waitlist(email_id=email_id, **waitlist.orm_dict())
+
+    attrs = waitlist.dict()
+    if "subscribed" in attrs:
+        del attrs["subscribed"]
```
grahamalama, line 483: Just want to make sure I understand what's happening here:

- `create_waitlist` is used in `create_contact` and `create_or_update_contact`
- It uses the `WaitlistInSchema`, which has a `subscribed` property 
  - this makes sense because we're receiving this payload from `Basket`, and waitlists are just newsletters with naming conventions, so it has this property
- we want to use this property to determine whether or not we should delete or create the waitlist for the contact, but our `Waitlist` ORM model currently doesn't have an update property, so we need to remove it

ctms/crud.py
```diff
@@ -477,11 +477,12 @@ def create_waitlist(
...
+
+    db_waitlist = Waitlist(email_id=email_id, **attrs)
```
grahamalama, line 485:
```suggestion
    db_waitlist = Waitlist(email_id=email_id, **waitlist.dict(exclude={"subscribed"}))
```

grahamalama (approved):
Looks good! Some of the testing methodology does pain me, but hopefully we'll continue to make improvements here as we improve our schemas

ctms/crud.py
leplatrem: Nice! Thanks!

ctms/crud.py
leplatrem: Yes, this is it.
I'll copy part of your note in a code comment

Issue #681, opened 2023-05-22
Enhance the Quality Standards score to exceed 80%


grahamalama:
- `monitoring-backends-duration` in #191 
- `monitoring-logging-env-configuration`
- `documentation-security-assessment`
- `documentation-troubleshooting`

leplatrem:
Done, missing checks listed in latest score card https://mozilla.github.io/syseng-pod/quality-standards/#service-CTMS

PR #718, opened 2023-06-27
Allow for more testing flexibility for local dev

grahamalama:
Move all logic required for running the test suite into the `bin/test.sh` script so that we have more flexibility for how we run the tests.

We can still run `make test`, which will run all of the tests and produce a coverage report like before. But now, we can _also_ run `bin/test.sh` to optionally enable the coverage report and supply `pytest` args for more flexibility in how we run the test suite for local development, e.g. `bin/test.sh -k "test_module"`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,13 +79,7 @@ start: .env
 
 .PHONY: test
 test: .env $(INSTALL_STAMP)
-	${DOCKER_COMPOSE} up --wait postgres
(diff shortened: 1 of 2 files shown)


leplatrem (approved):
Nice!

I was going to suggest to mention it in the docs but I saw the dev docs, and they would benefit from a big pruning... I'll cover it in #681

PR #719, opened 2023-06-27
Start to remove the usage of most contact fixtures from tests

grahamalama:
This PR will start a chain of (where appropriate) removing the use of `example_contact`, `maximal_contact`, `minimal_contact` and `most_minimal_contact` and replacing them with objects built with factories.

While we do this, we're also going to adjust test to be less brittle and have appropriate assertions.

This PR:

- Adds missing contact-related factories
- moves schema test modules into their own package
- Starts to adjust tests to remove the usage of our contact fixtures


leplatrem (approved):
nitpick: you replaced some contact lists `(example_contact, maximal_contact, minimal_contact)` by a single contact.
There maybe cases where the code executes a loop, and testing a single iteration may miss certain bugs (variable re-initialization within the loop, etc.).

grahamalama:
> There maybe cases where the code executes a loop, and testing a single iteration may miss certain bugs

I felt pretty confident about reducing those tests to just one contact. The tests in question are related to `PendingAcousticRecord`s. That model only relies on an `email_id`, and each contact, no matter how maximal or minimal, has an email id.

Looking back though, I do think it'd be valuable to add some test cases that cover `batch_limit` and `end_time` for 

```python
def get_all_acoustic_records_before(
    db: Session, end_time: datetime, retry_limit: int = 5, batch_limit=None
)```

grahamalama:
These acoustic test changes are getting pretty sizable -- I'm going to close this PR in favor of two

PR #492, opened 2022-12-14
 Fix #469: implement waitlists as generic relationship 

leplatrem:
See #470 

* [x] Introduce new models
* [x] Mimic what is done with newsletters
* [x] Fix tests
* [x] Test delete / replace waitlists (`DELETE` / unsubscribe ?)
* [x] Validate waitlist fields based on their name (currently only `vpn` since all `geo` is in the base schema)
* [x] Replace `vpn_waitlist`  by "dynamic" fields that will select in waitlist table
* [x] Process input data to replace `vpn_waitlist` fields in request body by entries in `waitlists` field
* [x] Migrate vpn records, create records in the waitlists table for contact with  `vpn_waitlist`
* [x] Process input data to replace `relay_waitlist` fields in request body by entries in `waitlists` field
* [x] Replace `relay_waitlist`  by "dynamic" fields that will select in waitlist table
* [x] Migrate relay, create records in the waitlists table for contact with `relay_waitlist`
* [x] Modify sync with Acoustic to read vpn and relay from waitlists list instead of fields
* [x] Figure out what to do with `sync_bq_tables`
* [x] Use something ~~like JSONPatch~~ newsletters to modify waitlists (instead of replacing the whole list of contact is field is specified in PATCH)
* [x] Verify correctness of data migration (eg. delete newsletters rows? migrate graceland into waitlists?)
* [x] Should `geo` be an explicit column or a field in the `fields`? (eg. does Graceland have geo?)
* [x] Add logging/metrics to track usage of legacy input format for vpn/relay
* [x] Create waitlists entries when subscribing to `graceland-waitlist`, `vpn-bundle-waitlist` newsletters?

PR #722, opened 2023-06-29
Ref #492, Fix #734: drop write support of legacy `vpn_waitlist` and `relay_waitlist` fields

### LINKED ISSUE #734: Unexpected behavior when `PUT`ing a contact with legacy waitlists
1. POST contact with a legacy waitlist field (VPN):
```
...
 'relay_waitlist': None,
 'vpn_waitlist': {'geo': 'us', 'platform': None},
 'waitlists': []
}
```
success ✅ 

2. GET that object back, looks like this:
```
 'relay_waitlist': {'geo': None},
...
 'vpn_waitlist': {'geo': 'us', 'platform': None},
 'waitlists': [{'fields': {'geo': 'us', 'platform': None},
                'name': 'vpn',
                'source': None}]
}
```
3. PUT the contact, signing up for another waitlist through a legacy field (Relay):
```
 'relay_waitlist': {'geo': 'us'},
 'vpn_waitlist': {'geo': 'us', 'platform': None},
 'waitlists': [{'fields': {'geo': 'us', 'platform': None},
                'name': 'vpn',
                'source': None,
                'subscribed': True}]
}
```

Expected: User is signed up for the relay and vpn waitlists
Actual: User is only signed up for the VPN waitlist

I'm pretty sure that's because of [this](https://github.com/mozilla-it/ctms-api/blob/82d2f70c9905d0797135b60355a2929c4a4d928b/ctms/backport_legacy_waitlists.py#L23-L25).

leplatrem:
Ref #492
Fix #734

* [x] Remove the code
* [ ] Make sure we don't receive legacy requests from Basket using STAGE and the `legacy_waitlists_requests` metric counter
* [ ] merge 🎉


grahamalama (approved):

leplatrem:
Screenshot 2023-07-13 at 19 19 40

Telemetry shows that some data is still provided in these fields. But this pull-requests shows, via the integration tests, that Basket version `2023-04-24` does not rely on those fields, otherwise the tests would fail. 

We have to investigate whether the metric is wrong, or whether some clients really rely on these fields

leplatrem:
@grahamalama I think we can merge this. 

Since Basket does not use the legacy format, and since MoFo does not manipulate waitlist subscriptions, I think we could tag a breaking change release and roll it out.

The metric is still increased but it is not well implemented, since our consumer do a read before a patch, and since the legacy fields are present in the read response, then the metric is always incremented on patch (`("vpn_waitlist" in formatted or "relay_waitlist" in formatted):` is true)

grahamalama:
Still looks good! I'm surprised our integration test didn't have to change at all.

Looks like this PR was focused on removing the option to create / update waitlists with the legacy format. Do we know if Basket still reads the [legacy waitlist output](https://github.com/mozilla-it/ctms-api/blob/6fdb18c9a6ad5dbd0a90181f6a84c6d1141af84b/ctms/schemas/contact.py#L294)? If not, maybe we can drop that too.

grahamalama (approved):
Still looks good! I'm surprised our integration test didn't have to change at all.

Looks like this PR was focused on removing the option to create / update waitlists with the legacy format. Do we know if Basket still reads the [legacy waitlist output](https://github.com/mozilla-it/ctms-api/blob/6fdb18c9a6ad5dbd0a90181f6a84c6d1141af84b/ctms/schemas/contact.py#L294)? If not, maybe we can drop that too.

leplatrem:
> Do we know if Basket still reads the [legacy waitlist output](https://github.com/mozilla-it/ctms-api/blob/6fdb18c9a6ad5dbd0a90181f6a84c6d1141af84b/ctms/schemas/contact.py#L294)?

It doesn't, and you're right we could definitely remove it

PR #723, opened 2023-06-29
Create mixin for create and update timestamp columns

grahamalama:
We were creating timestamps for row creation and update in a few different ways throughout our models. This commit aligns our approach in a mixin:

- uses `TIMESTAMP` because the SQLAlchemy docs[^1] suggest that if you're using `sa.DateTime` with `timezone=True`, you're better off using a `TIMESTAMP` instead

[^1]: https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.DateTime

- uses `server_default` instead of `default` consistently
- aligns on using `onupdate` to generate update timestamps. `server_onupdate` _would_ be the better approach, but Postgres doesn't support the `ON UPDATE` clause for regular columns. So, we delegate that responsibility to the ORM (consistently)


ctms/models.py
```diff
@@ -240,18 +222,8 @@ class ApiClient(Base):
...
-    update_timestamp = Column(
-        TIMESTAMP(timezone=True),
-        nullable=False,
-        server_default=now(),
-        server_onupdate=now(),
-    )
-
 
-class MozillaFoundationContact(Base):
+class MozillaFoundationContact(Base, TimestampMixin):
```
leplatrem, line 226: This model didn't have any but I guess it's better for consistency

ctms/crud.py
```diff
@@ -113,8 +113,9 @@ def get_all_contacts_from_ids(db, email_ids):
 
 def get_bulk_query(start_time, end_time, after_email_uuid, mofo_relevant):
     filters = [
+        # pylint: disable-next=comparison-with-callable
         Email.update_timestamp >= start_time,
-        Email.update_timestamp < end_time,
+        Email.update_timestamp < end_time,  # pylint: disable=comparison-with-callable
```
leplatrem, line 118: could have used `# pylint: disable-next=` here too...

migrations/versions/20230629_5bfd08b53489_create_update_timestamp_mixin.py
```diff
@@ -0,0 +1,149 @@
...
+        server_default=sa.text("now()"),
+        existing_nullable=False,
+    )
+    op.alter_column(
+        "pending_acoustic",
+        "update_timestamp",
+        existing_type=postgresql.TIMESTAMP(timezone=True),
+        server_default=sa.text("now()"),
+        existing_nullable=False,
+    )
```
leplatrem, line 78: I'm curious to see what SQL instruction comes out of this :)

leplatrem (approved):

ctms/models.py
grahamalama: Yeah that's what I thought as well. Why not have it.

PR #740, opened 2023-07-06
Fix #733: fix replacing the list of waitlists in PUT requests 

### LINKED ISSUE #733: Inconsistent handling of empty `waitlists` and `newsletters` lists when `PUT`ing a contact
I noticed an inconsistency in the way we treat the `newsletters` and `waitlists` fields of a contact when they're `PUT` to CTMS.

Consider a contact that we `GET` from CTMS:

```
 'newsletters': [{'format': 'T',
                  'lang': 'iu',
                  'name': 'newsletter-0',
                  'source': 'http://example.com/',
                  'subscribed': True,
                  'unsub_reason': None}],
 'waitlists': [{'fields': {},
                'name': 'waitlist-0',
                'source': 'https://example.com/'}]
```

then we `PUT` the contact back, but substitute the `newsletters` and `waitlists` fields with empty lists:

```
 'newsletters': [],
 'waitlists': []
```

When we `SELECT` that record from the database and look at associated rows in newsletters / contacts, we observe that:

there are no associated newsletters -- the row has been deleted
there _is_ still an associated waitlist -- the row has not been deleted

We should handle both of these cases consistently.

And, because of the reasons discussed in ADR 2, introduced in #707, we should probably mark the associated rows as `subscribed=False` instead of deleting them so that we can properly sync the contacts with Acoustic. 

Perhaps we only include `subscribed=True` newsletters and waitlists when `GET`ting the contact, though.

leplatrem:
Fix #733

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -487,6 +487,13 @@ def create_waitlist(
 def create_or_update_waitlists(
     db: Session, email_id: UUID4, waitlists: List[WaitlistInSchema]
 ):
+    names = [waitlist.name for waitlist in waitlists if not waitlist.is_default()]
+    db.query(Waitlist).filter(
+        Waitlist.email_id == email_id, Waitlist.name.notin_(names)
+    ).delete(
+        synchronize_session=False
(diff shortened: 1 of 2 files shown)


leplatrem:
Closing, needs a better approach

PR #744, opened 2023-07-06
Combine CI checks into one workflow

grahamalama:
I think this better-fits the "mental model" of Github actions workflows -- instead of one workflow file per job, each workflow file should represent a, well, workflow e.g. `deploy`, `ci`. That workflow may be composed of 1 or more jobs that run to fulfill that workflow.


grahamalama:
Because we have `run_test` as a required check, but now the job is called `run_unit_tests`, I think we'll have to bypass branch protections for this PR (if we decide to merge it), then update the branch protections accordingly

leplatrem (approved):
That could become a potential quality standard :) I take note of it

Issue #60, opened 2021-02-24
Add DELETE /ctms/{email_id} and soft deletion


jwhitlock:
Add a ``DELETE /ctms/{email_id}`` method to soft-delete a contact:

* Add a database field to ``email`` representing a soft-delete, or a soft-delete table
* When called, soft-delete the record, return ``410 Gone``
* ``GET /ctms/{email_id}`` returns a ``404 Not Found`` or ``410 Gone``
* ``POST /ctms`` or ``PUT /ctms/{email_id}`` is allowed without violating uniqueness or ID collisions

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-51)

jwhitlock:
Moving to Post-Salesforce milestone, since we're not planning to add a deletion process to the API before launch.

grahamalama:
Closing this since https://github.com/mozilla-it/ctms-api/pull/548 added a delete endpoint, however:
- It's a "hard delete" endpoint, not soft
- it's `DELETE /ctms/{primary_email}`, not `DELETE /ctms/{email_id}`

PR #745, opened 2023-07-06
Fix #733: fix replacing the list of waitlists in PUT requests

### LINKED ISSUE #733: Inconsistent handling of empty `waitlists` and `newsletters` lists when `PUT`ing a contact

leplatrem:
Fix #733 

I followed the *idempotent PUT/GET roundtrip* lead, and decided to delete existing waitlists that are not specified in a PUT request. This follows what was done for newsletters in https://github.com/mozilla-it/ctms-api/pull/112


ctms/crud.py
```diff
@@ -487,6 +490,17 @@ def create_waitlist(
...
 ):
+    # Start by deleting the existing waitlists that are not specified as input.
+    # We delete instead of set subscribed=False, because we want an idempotent
+    # round-trip of PUT/GET at the API level.
+    # Note: the contact is marked as pending synchronization at the API routers level.
+    names = [waitlist.name for waitlist in waitlists if not waitlist.is_default()]
+    db.query(Waitlist).filter(
+        Waitlist.email_id == email_id, Waitlist.name.notin_(names)
+    ).delete(
+        synchronize_session=False
```
grahamalama, line 501: This seems like it could throw a wrench into our waitlists ADR that we just added.

If we end up deleting waitlist rows as the result of a `PUT`, how will we propagate the deletions to Acoustic? It seems like for this to work, we would have needed to select [option 2](https://github.com/mozilla-it/ctms-api/blob/main/docs/adrs/waitlists_subscribed.md#option-2---reset-before-add). Same goes for newsletters.

Also, this begs the question -- even when a user unsubscribes from a newsletter (say with a `PATCH`), is the matching row deleted from the relational table?

tests/unit/routers/contacts/test_api.py
```diff
@@ -182,7 +191,6 @@ def test_post_get_put(client, post_contact, put_contact, update_fetched):
 
     # TODO: remove this once we remove support of `vpn_waitlist` and
     # `relay_waitlist` as input.
-    # If we don't strip these two fields before turning the data into
```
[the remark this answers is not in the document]
leplatrem: obviously not

ctms/crud.py
leplatrem: You're right, this deserves some clarification in the ADR.

> It seems like for this to work, we would have needed to select [option 2](https://github.com/mozilla-it/ctms-api/blob/main/docs/adrs/waitlists_subscribed.md#option-2---reset-before-add). Same goes for newsletters.

Option 2 is fragile and would not allow to store unsub reason.

Actually, it's a matter of what do we consistency with. And your comment points at a fundamental design flaw of CTMS: the fact that the CRUD REST API is tangled with the subscription API used by Basket.
 
Currently, one can delete contacts using a `DELETE` request, and everything is deleted from the DB, and no way for us to sync it to Acoustic (see #60).

This `PUT` request implementation is not different. Updates of list attributes in `PUT` requests are destructive. Except when an empty list is passed, the field is ignored (:facepalm).

`PATCH` do not behave like this (I know they shouldn't, that's why they is like https://jsonpatch.com for that).
They are a sort of hybrid with custom stuff for Basket.
In order to unsubscribe a user from all newsletters or waitlists, Basket sends either an `"UNSUBSCRIBE"` literal or a list of objects with `subscribed=false`.  The unspecified existing records are not deleted.

https://github.com/mozilla-it/ctms-api/blob/e51e21b9ad2ab464ff3f4c8e442675e978da8a33/ctms/crud.py#L618

[Basket does not use `DELETE` and `PUT`](https://github.com/search?q=repo%3Amozmeao%2Fbasket%20put%28&type=code).

I would say that we have several solutions at hand:

* Leave it like this PR. Consistency between newsletters and waitlists in PUT requests. And cross fingers so that no consumer uses the `PUT` and `DELETE` with expectation to see the changes propagated to Acoustic.

* Change the spec, and implement the same soft deletion in `PUT` for list attributes. Consistency between PUT and PATCH.

* Restrict the destructive endpoints that can lead to inconsistent state in Acoustic. Only allow endpoints whose changes can be sync'ed in Acoustic. Consistency of sync expectations. We may not have to implement a full ACL system, maybe just a flag on `ApiClient` for admin operations like DELETE.

* Remove the `PUT` endpoint completely. My favorite, but does not solve the situation for `DELETE` which is currently open any token holder

grahamalama (approved):

PR #746, opened 2023-07-06
Add Github Actions to dependabot checks

grahamalama:


leplatrem (approved):
Would also be a candidate for quality standards 😎 (took note)

PR #760, opened 2023-07-13
Untangle simplify and fix heartbeat code 

leplatrem:
While working on #681 I realized the heartbeat code was funky.

- `monitor.py` contained `get_version()` (?)
- the check database was in charge of passing settings values
- heartbeat would return 200 even if the acoustic would fail
- the threshold comparisons were inverted and associated tests did not catch it because of bad assignments


ctms/crud.py
```diff
@@ -57,6 +58,10 @@
 from .schemas.base import BaseModel
 
 
+def ping(db: Session):
+    db.execute(select([func.now()])).first()
```
grahamalama, line 62: What if we did our exception capture here?

```suggestion
def ping(db: Session):
    try:
        with db.connection():
            return True
    except Exception as exc:
        logger.exception(exc)
        return False
```

then we can safely do this below:

```python
alive = ping(db)
```

ctms/routers/platform.py
```diff
@@ -128,29 +129,62 @@ def heartbeat(
...
+            ("backlog", backlog, settings.acoustic_max_backlog),
+            ("retry backlog", retry_backlog, settings.acoustic_max_retry_backlog),
+        ]:
+            if value is not None and maximum is not None and value > maximum:
+                logger.error(
+                    f"Acoustic {name} size %s exceed maximum %s",
+                    backlog,
+                    settings.acoustic_max_backlog,
+                )
+                acoustic_success = False
```
grahamalama, line 172: This feels like it could be a separate function, no?

grahamalama (approved):

PR #768, opened 2023-07-17
Use timestamps instead of dates in Acoustic records

leplatrem:
This is a drive-by that I found out while working on #762
There were two contradictory code comments. 

From what I understand in the docs, the queries don't take the time part into account, but timestamps can be stored.
https://help.goacoustic.com/hc/en-us/articles/360043347833-Queries-and-time-stamp-fields

```
Timestamp | Allows date and time in format mm/dd/yyyy hh:mm:ss
```

**Update**

Trying to store a date in timestamp column fails actually:

https://mozilla.sentry.io/issues/4320526677/?referrer=slack&alert_rule_id=13795152&alert_type=issue

So this becomes a bug fix.

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -68,12 +68,9 @@ def force_bytes(s, encoding="utf-8", strings_only=False, errors="strict"):
 def transform_field_for_acoustic(data):
     """Transform data type for Acoustic."""
     if isinstance(data, bool):
-        if data:
-            return "Yes"
-        return "No"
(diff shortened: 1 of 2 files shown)


leplatrem:
**Update**

Trying to store a date in timestamp column fails actually:

https://mozilla.sentry.io/issues/4320526677/?referrer=slack&alert_rule_id=13795152&alert_type=issue

So this becomes a bug fix.

ahoneiser (approved):

grahamalama (approved):

PR #770, opened 2023-07-19
Remove notion of `trivial` logs

grahamalama:
in PR #210, we introduced the notion of "trivial" requests to allow us to more-quickly filter logs in Papertrail for logs that actually mattered. This was partially due to the fact that Papertrail had poor filtering capabilities.

Now, since we're using a better logging platform, we can remove this "trivial" flag for logs and use Google Cloud's superior filtering capabilities to do our filtering.


leplatrem (approved):
🤩

PR #804, opened 2023-08-30
Fix special boolean text values on the main table

grahamalama:
Ref #802

PR #805, opened 2023-08-31
Fix #802: clarify sync behaviour of boolean fields for relations 

### LINKED ISSUE #802: Column double opt-in receives `No`/`Yes` instead of `0`/`1`
Since #762 the boolean of the main contact table are synchronized to Acoustic with values `Yes`/`No` instead of `1`/`0`.

This is problematic since marketing relies on `0` and `1` text values to filter double opt-in and opted-out contacts.

leplatrem:
In #804 we fixed the way boolean fields are synced on the main table. This PR is a complementary specification about the way boolean related to FxA and AMO are synced.

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -259,8 +259,10 @@ class CTMSToAcousticService:
                             )
                         # TODO: These are boolean values on our models, but in Acoustic
                         # they are configured as text values where we record `True`
-                        # and `False` as `1` and `0` respectively. We should
-                        # configure Acoustic to use a boolean column for this data.
+                        # and `False` as `1` and `0` respectively.
+                        # Note that `amo_*` and `fxa_*` fields are omitted when contact does not
+                        # have these relations.
+                        # We should configure Acoustic to use a boolean column for this data.
                         # This is being tracked in https://github.com/mozilla-it/ctms-api/issues/803
                         if acoustic_field_name in (
                             "double_opt_in",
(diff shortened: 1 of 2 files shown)


ctms/acoustic_service.py
```diff
@@ -259,8 +259,10 @@ def _main_table_converter(self, contact, main_fields):
                             )
                         # TODO: These are boolean values on our models, but in Acoustic
                         # they are configured as text values where we record `True`
-                        # and `False` as `1` and `0` respectively. We should
-                        # configure Acoustic to use a boolean column for this data.
+                        # and `False` as `1` and `0` respectively.
+                        # Note that `amo_*` and `fxa_*` fields are omitted when contact does not
+                        # have these relations.
```
grahamalama, line 264: Should we mention explicitly that those values will be `null`? For example, we _could_ have written this code so that if the contact didn't have a relation, we'd record the acoustic value as `0`.

I think that's technically new behavior too for the main table columns. As we saw when looking in Acoustic yesterday, `double_opt_in` and `has_opted_out_of_email` were `null` in some places.

leplatrem: They will be `null` because no default is configured on the field in Acoustic. But someone could change the default to `no` and the code comment becomes obsolete. 
Screenshot 2023-08-31 at 15 05 18

Shall we mention that it relies on the Acoustic field default value? Is it worth it?

The subfields `amo` and `fxa` are skipped above in the iteration https://github.com/mozilla-it/ctms-api/blob/650d68e9b0fcbf2e83c8abcf41caa2bdf52e8568/ctms/acoustic_service.py#L237 
Without more changes, we can't set `0` there.

PR #811, opened 2023-09-05
Fix #806: reset retry and last_error columns on resync command

### LINKED ISSUE #806: `acoustic.py resync` should set `retries` column to `0`
If a contact has `retries=5`, it won't be synchronized.

If this contact is among the specified list for `acoustic.py resync` then it won't be resynced.

https://github.com/mozilla-it/ctms-api/blob/bd8b23d99efb6e4bb0523a66e7efd415236e2b44/ctms/crud.py#L334

We should upsert the `PendingAcousticRecord` with `retries=0`

leplatrem:
Fix #806

### DIFF
--- a/tests/unit/bin/test_acoustic.py
+++ b/tests/unit/bin/test_acoustic.py
@@ -33,6 +33,23 @@ def test_main_force_resync_by_email_list(dbsession, maximal_contact, tmpdir):
     assert len(dbsession.query(PendingAcousticRecord).all()) > 0
 
 
+def test_main_force_resync_will_reset_existing_retries(
+    dbsession, maximal_contact, tmpdir
+):
+    record = PendingAcousticRecord(email_id=maximal_contact.email.email_id, retry=99)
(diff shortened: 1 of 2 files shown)


tests/unit/bin/test_acoustic.py
```diff
@@ -33,6 +33,23 @@ def test_main_force_resync_by_email_list(dbsession, maximal_contact, tmpdir):
     assert len(dbsession.query(PendingAcousticRecord).all()) > 0
 
 
+def test_main_force_resync_will_reset_existing_retries(
```
grahamalama, line 36: Based on my [comment here](https://github.com/mozilla-it/ctms-api/issues/806#issuecomment-1716282544), should we add a test that demonstrates when there are two pending records for the same `email_id`, after running `do_sync`, only one will remain with `retry=0`?

Or we could land the fix for duplicate email_ids in the queue in this PR.

Issue #803, opened 2023-08-30
Migrate main contacts table boolean fields to boolean type


leplatrem:
These fields are currently text fields:

- `double_opt_in`
- `has_opted_out_of_email`
- `email_opt_in`
- `account_deleted`

that receive 0 or 1.

And could be proper booleans (like waitlist subcribed column)

This migration could be part of the main table schema changes of #563

Related #802

PR #813, opened 2023-09-07
Fix #812: fix Acoustic create_timestamp column

### LINKED ISSUE #812: Sync a contact's `create_timestamp` to Acoustic as an Acoustic date
https://help.goacoustic.com/hc/en-us/articles/360043349433-Work-with-database-fields

We currently sync a contact's `create_timestamp` to Acoustic in a format that is incompatible with Acoustic's `Date` type. As @leplatrem discovered:

currently on `main`
```
poetry run ctms/bin/acoustic.py dump --query="SELECT email_id FROM emails LIMIT 1;"

Dump CSV for 1 contacts [y/N]? y
...
...05/21/2021 16:39:02,...
```

from `v2.1.0`
```
poetry run ctms/bin/acoustic.py dump --query="SELECT email_id FROM emails LIMIT 1;"

Dump CSV for 1 contacts [y/N]? y
...
...05/21/2021,...
```

leplatrem:
The regression introduced in https://github.com/mozilla-it/ctms-api/commit/4ee09507cc461bcf5e4538dd167203a8ac8fe43d was not spotted because no unit test was asserting the output format of the `create_timestamp` format.

Also, Acoustic was ignoring silently the value sent in the `create_timestamp` column instead of rejecting the row. 

Note that we could have expected that sending `"05/21/2021 16:39:02"` for a `Date` field would have been ok...

### DIFF
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -256,7 +256,7 @@ class CTMSToAcousticService:
                         # TODO: The create_timestamp is currently a Date field in Acoustic.
                         # Sending time information will prevent Acoustic to set the value properly.
                         # We should configure Acoustic to use a Timestamp for this data.
-                        # This is being tracked in https://github.com/mozilla-it/ctms-api/issues/803
+                        # This is being tracked in https://github.com/mozilla-it/ctms-api/issues/563
                         if acoustic_field_name == "create_timestamp":
                             inner_value = inner_value.date()
 
--- a/ctms/acoustic_service.py
+++ b/ctms/acoustic_service.py
@@ -253,6 +253,13 @@ class CTMSToAcousticService:
                         self.context["trace"] = inner_value
 
                     if acoustic_field_name in main_fields:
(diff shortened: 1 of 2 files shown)


ctms/acoustic_service.py
```diff
@@ -253,6 +253,13 @@ def _main_table_converter(self, contact, main_fields):
                         self.context["trace"] = inner_value
 
                     if acoustic_field_name in main_fields:
+                        # TODO: The create_timestamp is currently a Date field in Acoustic.
+                        # Sending time information will prevent Acoustic to set the value properly.
+                        # We should configure Acoustic to use a Timestamp for this data.
+                        # This is being tracked in https://github.com/mozilla-it/ctms-api/issues/803
```
grahamalama, line 259: This is the issue for migrating "boolean-like" columns to actual booleans.

Should we make a new issue or expand the scope of #803 to track migrating all "date-like" columns to dates or timestamps?

grahamalama (approved):

PR #816, opened 2023-09-08
Fix #814: add timestamps to newsletters and waitlists in contact responses

### LINKED ISSUE #814: Expose `create_timestamp` and `update_timestamp` of newsletters in contact response
See [this doc](https://docs.google.com/document/d/1WqkGu_TLPwwKGGDwBzcm3yACsWTVolkVsuKP0gdSpoI/edit#heading=h.kz4f4stt03px)

These seem to be the only API changes required for this integration.

leplatrem:
I'm not a big fan for those `del actual["create_timestamp"]` in the tests.

We could either have a `omit(dict, keys="")` util, or, better IMO, we could have a test util to fuzzy match values instead of actual timestamp values.

For example:

https://github.com/mozilla/normandy/blob/8ec2b525493ee0af7da3e8beee4be4f1669f740c/normandy/base/tests/__init__.py#L12-L51

```py
        assert serializer.data == {
            "signature": Whatever.regex(r"[a-f0-9]{40}"),
            "x5u": Whatever.startswith(signature.x5u),
            "timestamp": Whatever.iso8601(),
            "public_key": Whatever.regex(r"[a-zA-Z0-9/+]{160}"),
        }
```
https://github.com/mozilla/normandy/blob/8ec2b525493ee0af7da3e8beee4be4f1669f740c/normandy/recipes/tests/api/v1/test_serializers.py#L140-L145

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,19 +4,17 @@ Common test configuration both unit and integration tests.
 
 from datetime import datetime
 
-import pytest
 
-
-class Whatever:
+class FuzzyAssert:
     """
     This class is a testing helper that provides flexible equality
     of values.
 
     .. code-block::
 
-        >>> Whatever(lambda x: x.startswith("a")) == "abc"
+        >>> FuzzyAssert(lambda x: x.startswith("a")) == "abc"
         True
-        >>> Whatever(lambda x: x % 2 == 0) == 11
+        >>> FuzzyAssert(lambda x: x % 2 == 0) == 11
         False
 
     It is mainly used to make sure fields contain valid dates
@@ -24,9 +22,9 @@ class Whatever:
 
     .. code-block::
 
-        >>> Whatever.iso8601() == "2020-01-01"
+        >>> FuzzyAssert.iso8601() == "2020-01-01"
         True
-        >>> Whatever.iso8601() == None
+        >>> FuzzyAssert.iso8601() == None
         False
     """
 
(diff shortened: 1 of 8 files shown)


leplatrem:
@grahamalama r?

tests/conftest.py
```diff
@@ -0,0 +1,37 @@
+"""
+Common test configuration both unit and integration tests.
+"""
+
+from datetime import datetime
+
+import pytest
+
+
+class Whatever:
```
grahamalama, line 10: Can we add some documentation here for how this class works? The end result seems useful, but I'm not quite sure how it works at first glance. Also, perhaps it deserves a more descriptive name?

grahamalama (approved):

tests/conftest.py
leplatrem: I added a docstring. I took the name from the Normandy source code. I'm happy to rename if you have suggestions!

grahamalama: What about something with the word "constrain", maybe `Constrainer` or `ConstrainedAssertion`?

leplatrem: Oh that's interesting. I see it as the complete opposite, like *loose assertion*, *sloppy assertion*

grahamalama: As we ended up discussion on Slack, we landed on something to do with `Fuzzy`, like `FuzzyAssertion`.

PR #892, opened 2024-03-12
Turn suppression list CSV into a set of SQL files to be ran on the server 

leplatrem:
* [x] polish expectations on input CSV format (eg. header names, double quotes)
* [x] ~estimate time of execution of join query on primary_email in prod~ turn joins into batches

### DIFF
--- a/suppression-list/csv2optout.py
+++ b/suppression-list/csv2optout.py
@@ -79,7 +79,7 @@ UPDATE emails
 
 INSERT INTO pending_acoustic(email_id, retry)
   SELECT email_id, 0 FROM optouts_{tmp_suffix}
-WHERE idx > {start_idx} AND idx <= {end_idx};
+WHERE {schedule_sync} AND idx > {start_idx} AND idx <= {end_idx};
 
 DELETE FROM optouts_{tmp_suffix}
   WHERE idx > {start_idx} AND idx <= {end_idx};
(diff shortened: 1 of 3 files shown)


suppression-list/csv2optout.py
```diff
@@ -0,0 +1,208 @@
...
+    # Prepare SQL files
+    #
+    now = datetime.now(tz=timezone.utc)
+    tmp_suffix = now.strftime("%Y%m%dT%H%M")
+    batch_commands = []
+    for i in range(batch_count):
+        start_idx = i * batch_size
+        end_idx = (i + 1) * batch_size
+        batch_commands.append(
+            SQL_BATCH.format(
```
alexcottner, line 164: Possible injection flaw. This is a one-off and we aren't going to reuse it in application code, so not gonna push hard. But ideally this would be a parameterized query or database function.

suppression-list/csv2optout.py
```diff
@@ -0,0 +1,208 @@
...
+
+SQL_COMMANDS_PRE = """
+CREATE OR REPLACE PROCEDURE raise_notice (s TEXT) LANGUAGE plpgsql AS
+$$
+BEGIN
+    RAISE NOTICE '%', s;
+END;
+$$;
+
+CREATE TEMPORARY TABLE IF NOT EXISTS imported_{tmp_suffix} (
```
alexcottner, line 21: Why do we need a table suffix for a temp table? The table would be dropped at the end of the session by default (or at the end of a transaction if we specify `on commit drop`)

suppression-list/csv2optout.py
```diff
@@ -0,0 +1,208 @@
...
+  FROM '{csv_path}' -- fullpath
+  DELIMITER '{delimiter}'
+  {headers}
+  QUOTE AS '"';
+
+CALL raise_notice('CSV import done.');
+
+CALL raise_notice('Join on existing contacts...');
+
+CREATE TABLE IF NOT EXISTS optouts_{tmp_suffix} (
```
alexcottner, line 39: We could make this one a temp table too and let it drop at session close. That would save us from cleanup and remove the need for a temp suffix.

leplatrem: I don't think so, the idea is that the update batches can be executed over the course of several days if necessary

suppression-list/csv2optout.py
leplatrem: Yes, you're right, this is not necessary

suppression-list/csv2optout.py
leplatrem: Yeah, well, the script outputs a sql file that is meant to be executed in `psql`

alexcottner (approved):
seems reasonable for a one-off import

leplatrem:
I'll merge this because it was done. If the location in the repo is not adecuate, I'm happy to move it.

PR #903, opened 2024-05-15
Remove Stripe integration / product data

grahamalama:
We previously ingested Stripe data from FXA in order to associate paid products with users. This was eventually pushed to Acoustic as a relational table. Now, since we have the option of joining this data in Bigquery before pushing to Braze, we no longer require this integration.

We remove the infrastructure to push this data here to us here:
https://github.com/mozilla-services/cloudops-infra/pull/5658
https://github.com/mozilla-services/cloudops-infra/pull/5659

When we merge this PR, on the stage infra will have been torn down. When this commit is included in a release, the prod infra will have been torn down


leplatrem (commented):
This seems reasonable indeed.

leplatrem (approved):

PR #919, opened 2024-06-10
Remove import side effects from `database.py`

leplatrem:


ctms/dependencies.py
```diff
@@ -19,6 +19,7 @@ def get_settings() -> Settings:
 
 
 def get_db():  # pragma: no cover
+    SessionLocal = session_factory()
```
grahamalama, line 22:
```python
def session_factory(scoped: bool = False):
    factory = sessionmaker(autocommit=False, autoflush=False, bind=get_engine())
```

Does this mean we're going to create a new engine every time a route fetches a DB connection with this dependency? Something from the SQLAlchemy [docs](https://docs.sqlalchemy.org/en/14/core/connections.html):

> The typical usage of [create_engine()](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine) is once per particular database URL, held globally for the lifetime of a single application process. A single [Engine](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Engine) manages many individual [DBAPI](https://docs.sqlalchemy.org/en/14/glossary.html#term-DBAPI) connections on behalf of the process and is intended to be called upon in a concurrent fashion.

tests/unit/conftest.py
```diff
@@ -123,7 +122,8 @@ def engine(pytestconfig):
 def connection(engine):
     """Return a connection to the database that rolls back automatically."""
     conn = engine.connect()
-    SessionLocal.configure(bind=conn)
+    factories.SessionLocal.configure(bind=conn)
+    factories.ScopedSessionLocal.configure(bind=conn)
```
grahamalama, line 126: Why do we need both `SessionLocal` and `ScopedSessionLocal` here?

grahamalama (commented):
With this PR's current direction, I'm concerned that we're straying away from some of SQLAlchemy's best practices when it comes to managing connections -- mainly, the bit about creating a global engine object.

> to have custom (simplified) DB settings

If this is the goal, maybe we can do some tuning here? Perhaps we can make it so the settings that are required by our database module are separate from CTMS's other settings. So we still create a global engine object that requires some (minimal) database settings where we can provide defaults that can be overridden with Pydantic

```python

class DBSettings(BaseSettings):
    url: PostgresDsn
    pool_size: int = 5  # Default value from sqlalchemy
    max_overflow: int = 10  # Default value from sqlalchemy
    pool_timeout_in_seconds: int = 30  # Default value from sqlalchemy
    pool_recycle_in_seconds: int = 900  # 15 minutes

    class Config:
        env_prefix = "ctms_db_"

engine = engine_factory(DBSettings())
```

But

leplatrem:
Thanks Graham for your detailed review 🎩

Your doubts convinced me that changing this part was too risky compared to the added value.

I followed your suggestion to split the settings class in https://github.com/mozilla-it/ctms-api/pull/929

PR #918, opened 2024-06-10
Fix #915: Use DB instead of API for delete_bulk

leplatrem:
Fix #915

* [x] Implement command using the ORM
* [ ] Remove requirement of the `CTMS_SECRET_KEY` env var

Currently, there's a side effect of importing `crud.py`, which instantiates `Settings()`
```
Traceback (most recent call last):
  File "/Users/mathieu/Code/Mozilla/ctms-api/ctms/bin/delete_bulk.py", line 7, in <module>
    from ctms.crud import delete_contact, get_contacts_by_any_id
  File "/Users/mathieu/Code/Mozilla/ctms-api/ctms/crud.py", line 15, in <module>
    from .backport_legacy_waitlists import format_legacy_vpn_relay_waitlist_input
  File "/Users/mathieu/Code/Mozilla/ctms-api/ctms/backport_legacy_waitlists.py", line 7, in <module>
    from .models import Waitlist
  File "/Users/mathieu/Code/Mozilla/ctms-api/ctms/models.py", line 21, in <module>
    from .database import Base
  File "/Users/mathieu/Code/Mozilla/ctms-api/ctms/database.py", line 18, in <module>
    engine = engine_factory(Settings())
                            ^^^^^^^^^^
  File "/Users/mathieu/Code/Mozilla/ctms-api/.venv/lib/python3.12/site-packages/pydantic/env_settings.py", line 40, in __init__
    super().__init__(
  File "/Users/mathieu/Code/Mozilla/ctms-api/.venv/lib/python3.12/site-packages/pydantic/main.py", line 341, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 2 validation errors for Settings
db_url
  field required (type=value_error.missing)
secret_key
  field required (type=value_error.missing)
```

Should we implement #919 or just set a default value to `secret` (eg. `""`), since it is already set in prod anyway?

Issue #915, opened 2024-06-07
Rewrite bulk delete command to use DB queries instead of API calls


leplatrem:
This would help us deprecate the DELETE endpoint, as part of the project sunset

Related #60

leplatrem:
This issue does not make sense anymore.

PR #929, opened 2024-06-27
Split DB from application settings

leplatrem:
This would allow commands to only read DB settings instead of redefining settings class or requiring unused config values like API keys.

Note: should we have a single `Settings` class that inherit from `DBSettings` and `AppSettings` or are we good with dealing with these two individually? 

See #918


leplatrem:
dropping since #915 was abandonned

Issue #549, opened 2023-02-15
Adopt a common pattern for commiting database changes


grahamalama:
Right now, we commit database transactions in many different places:
- `crud.py`
- `app.py`
- various scripts

We should adopt a common pattern for committing session data so that it all happens in one place.

leplatrem:
I would recommend:
- manual commit/rollbacks only in CLI scripts
- a middleware that commits if the response is 2XX and aborts if > 3XX
And nowhere else

PR #946, opened 2024-07-29
Refactor some canned fixtures for tests (minmal/maximal contacts, example contact)

grahamalama:
This PR takes a big swing at removing the usage of ambiguous canned fixutures in tests, namely:

- `most_minimal_contact`
- `minimal_contact`
- `maximal_contact`
- `example_contact`

I say ambiguous because at a glance, it's hard to know what the differences are between these fixtures.

Instead, we use factories to build models and commit those models to the database as needed in each test. This results in sometimes more verbose, but (in my opinion) clearer tests.

These fixtures aren't totally removed yet. In future PRs, we'll have to untangle them from how they're used in parametrized tests and other complex fixtures (e.g. `{put, post}_contact`)

### DIFF
--- a/tests/factories/models.py
+++ b/tests/factories/models.py
@@ -93,7 +93,7 @@ class AmoAccountFactory(BaseSQLAlchemyModelFactory):
     last_login = factory.Faker("date_object")
     location = factory.Faker("city")
     profile_url = factory.LazyAttribute(
-        lambda obj: f"https://www.example.com/{obj.display_name}"
+        lambda obj: f"https://ex.com/{obj.display_name}",
     )
     user = True
     user_id = factory.Faker("pystr", max_chars=40)
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -199,18 +199,6 @@ def create_full_contact(db, contact: ContactSchema):
 
 
 @pytest.fixture
-def most_minimal_contact(dbsession):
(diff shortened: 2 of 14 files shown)


leplatrem (approved):
I scanned the changes and it looks sound ✅ 

Since we `dbsession.commit()` everywhere where factories are used, wouldn't it simpler to do it in the factory function? 

Related #549

grahamalama:
> Since we dbsession.commit() everywhere where factories are used, wouldn't it simpler to do it in the factory function?

Good call here. This is a [factoryboy option](https://factoryboy.readthedocs.io/en/stable/orms.html#factory.alchemy.SQLAlchemyOptions.sqlalchemy_session_persistence) (`sqlalchemy_session_persistence = commit`), though I tried adopting that in this PR and I got some odd errors. Would be good for a follow up PR I think, since it might warrant some more investigation.

tests/factories/models.py
```diff
@@ -93,7 +93,7 @@ class Meta:
     last_login = factory.Faker("date_object")
     location = factory.Faker("city")
     profile_url = factory.LazyAttribute(
-        lambda obj: f"https://www.example.com/{obj.display_name}"
+        lambda obj: f"https://ex.com/{obj.display_name}",
```
grahamalama, line 96: This was generating URLS that were greater than our allowed 40 characters configured on the column

Issue #201, opened 2021-05-11
Sphinx - Auto Documentation


bsieber-mozilla:
Currently in the GitHub workflows, we have a mistake; which has prevented it from running and providing value.

https://github.com/mozilla-it/ctms-api/blob/7c035ae299956dab47ea020cbcc1aec84578b5aa/.github/workflows/sphinx-pages.yaml#L5

Changing this to `main` resulted in the workflow actually running, and failing:

https://github.com/mozilla-it/ctms-api/runs/2551274455?check_suite_focus=true

In live, sphinx auto-docs are not working. 

They are able to be produced and could be presented once the build is fixed.

![Screen Shot 2021-05-10 at 6 06 22 PM](https://user-images.githubusercontent.com/53873986/117745630-b34d7580-b1bf-11eb-9a9b-dc3722e924e9.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-36)

jwhitlock:
If we don't want to host them, we could build and host on https://readthedocs.org. The changes would be:

- Convert the guides to Sphinx
- Add autodocs
- Add an RTD web hook to the repo ([see guide](https://docs.readthedocs.io/en/stable/connected-accounts.html))

bsieber-mozilla:
I like that approach!

grahamalama:
Since we no longer use Sphinx to build documentation, we'll close this issue.

PR #955, opened 2024-08-05
Remove Sphinx from dev dependencies

leplatrem:
We don't Sphinx in this project... I was surprised to see #953


grahamalama (approved):
Nice. Apparently we did long ago:

https://github.com/mozilla-it/ctms-api/issues/201
https://github.com/mozilla-it/ctms-api/pull/266#issuecomment-932507279
https://github.com/mozilla-it/ctms-api/pull/304

But yeah let's get this out of here 🗑️

PR #958, opened 2024-08-08
Rewrite `POST` and `PUT` tests

grahamalama:
Rewrite `POST` and `PUT` tests so that they're more "straightforward" with less indirection.

Before, each test was instrumented like (in the `POST` tests):

`SAMPLE_CONTACT_PARAMS` ([src](https://github.com/mozilla-it/ctms-api/blob/main/tests/unit/conftest.py#L50))
`post_contact fixture` ([src](https://github.com/mozilla-it/ctms-api/blob/e2e43803b1fd1e508a6aa1b8637842e2f0cabf8a/tests/unit/conftest.py#L585))

```python
POST_TEST_PARAMS = pytest.mark.parametrize(
	"post_contact", SAMPLE_CONTACT_PARAMS, indirect=True
)


@POST_TEST_PARAMS
def test_create_basic_no_id(post_contact):
    ...
```

`PUT` tests were instrumented in the same way.

Now, tests are more straightforward in that they:

- (optionally) load the database with contact data, which is clearly evident in the test setup
- `POST` / `PUT` contact data, which is clearly evident in the test setup
- make assertions about what is created / updated / deleted and what is returned from the REST API

The `post_contact` / `put_contact` fixtures are now only used in one final test, but it is complex enough that it deserves its own PR.


leplatrem (approved):
Great 💯 
Glad we won't have this `_compare_written_contacts()` thing

PR #959, opened 2024-08-08
Leverage Dockerflow request summary

leplatrem:
* [x] basic wiring
* [x] make decision for extra fields around auth, etc.
       - emit other log line
       - add them to request.summary using subclass of dockerflow middleware
       - remove them

**After this PR**
- Requests logs are done in `request.summary` 
- `ctms.web` logs are gone
- headers, query params, and remote host are not logged anymore
-  path_template are not logged anymore (eg. `/ctms/{email_id}` instead of `/ctms/value`), but are maintained in metrics


tests/unit/test_log.py
```diff
@@ -25,32 +27,18 @@ def test_request_log(client, email_factory, caplog):
...
+        "agent": "testclient",
         "method": "GET",
         "path": f"/ctms/{email_id}",
-        "path_params": {"email_id": email_id},
-        "path_template": "/ctms/{email_id}",
-        "status_code": 200,
+        "code": 200,
+        "lang": None,
         "rid": "foo-bar",
+        "t": FuzzyAssert(lambda x: x > 0, name="uint"),
```
grahamalama, line 41: Any reason in particular you didn't want to pull the duration directly off the log object?

grahamalama (approved):
Looks good! 2 nits, but g2g if you'd like.

tests/unit/test_log.py
leplatrem: Like `"t": log.duration` ?

It was a way to really test the value. Like if `log.duration` is None then the assertion will be true

ctms/log.py
```diff
@@ -46,6 +63,7 @@
         "": {"handlers": ["console"]},
         "request.summary": {"level": logging.INFO},
         "ctms": {"level": logging.DEBUG},
+        "httpx": {"level": logging.CRITICAL},
```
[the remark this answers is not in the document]
leplatrem: Good call

tests/unit/test_log.py
grahamalama: Ahh right 🤦🏻

PR #975, opened 2024-09-27
Remove more fixtures

grahamalama:
This PR edits a few more tests so that we can remove the minimal / maximal contact fixtures.

The data that was used to create those fixtures is still used in one test, but in a follow-up PR we'll do that final cleanup.


tests/unit/test_crud.py
```diff
@@ -226,44 +243,30 @@ def test_get_contact_by_any_id_missing(dbsession, sample_contacts):
...
-            basket_token=str(uuid4()),
-            sfdc_id=(
-                alt_id_value if alt_id_name == "sfdc_id" else "other_sdfc_alt_id_value"
-            ),
-        ),
+    """Two contacts can share the same:
+    - fxa primary_email
+    - amo user_id
+    - sfdc_id
+    - mofo_contact_id
```
grahamalama, line 252: This was interesting to discover.

tests/unit/routers/contacts/test_api.py
```diff
@@ -176,6 +175,17 @@ def find_default_fields(contact: ContactSchema) -> Set[str]:
...
+    ("maximal_contact_data", set()),
+    ("example_contact_data", set()),
+    ("to_add_contact_data", set()),
+    ("simple_default_contact_data", {"amo"}),
+    ("default_newsletter_contact_data", {"newsletters"}),
+    ("default_waitlist_contact_data", {"waitlists"}),
+]
+
+
 @pytest.mark.parametrize("post_contact", SAMPLE_CONTACT_PARAMS, indirect=True)
```
leplatrem, line 189: Could even be put inline instead of using a constant

tests/unit/test_crud.py
leplatrem: Oh, that means there is no unique constraint on the main email table for this tuple of fields?
Do we know if we have occurences in prod data?

tests/unit/routers/contacts/test_api_get.py
```diff
@@ -74,171 +74,6 @@ def test_get_ctms_for_minimal_contact(client, email_factory):
...
-                "fields": {"geo": "ca", "platform": "windows,android"},
-                "name": "vpn",
-                "source": None,
-                "subscribed": True,
-                "unsub_reason": None,
-                "create_timestamp": "2010-01-01T08:04:00+00:00",
-                "update_timestamp": "2020-01-28T14:50:00+00:00",
-            },
-        ],
-    }
```
leplatrem, line 239: 🎉

leplatrem (approved):
Awesome Graham! 💯 Congrats for your patience on this.
I reviewed per commit and all seems sound 👌

tests/unit/test_crud.py
grahamalama: Yeah, I'm wondering why they were omitted, but even before, we had tests to assert that this behavior was true, so I'm wondering why it was modeled this way 🤔. Maybe it was intentional?

tests/unit/routers/contacts/test_api.py
grahamalama: True, but since it's (hopefully) being deleted very soon, I'll leave it as is in this PR.

PR #979, opened 2024-10-01
Fix #238: migration to backport prod index

### LINKED ISSUE #238: Add an alembic migration to sync with production indexes
There were indexes manually added to the staging and production deployments that also need to be added to the models and via an alembic migration. The migration should also remove any temporary indexes.

For example, this index was added:

```
CREATE UNIQUE INDEX idx_email_primary_unique_email_lower ON public.emails USING btree (lower((primary_email)::text));
```

There should be liberal use of ``IF EXISTS`` clauses, to avoid errors if an index already exists.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CTMS-26)

leplatrem:
Fix #238


grahamalama:
Trying to get the timeline correct here:

- bsieber-mozilla merged [this commit](https://github.com/mozilla-it/ctms-api/commit/b53fb58ca9f74625f2f7426a02c0d3e11842bd86#diff-4aa7c2a69c18e1f8b33b7cb5f54efdea916ea029661c212d3b1e4de349c91a24R76) on May 19, 2021, which created [this index](https://github.com/mozilla-it/ctms-api/commit/b53fb58ca9f74625f2f7426a02c0d3e11842bd86#diff-1a390c7b8c5ef6d2a2186016146f3b93ce3a2c04874ab1bfe02e8f67f350e8bfR25-R30)

```python
    op.create_index(
        "idx_email_primary_email_lower",
        "emails",
        [sa.text("lower(primary_email)")],
        unique=True,
    )
```

- on May 20, 2021, he merged [this commit](https://github.com/mozilla-it/ctms-api/commit/2036559ded34d78d2c2418840e8362df46a06ce8), which [removed the unique constraint](https://github.com/mozilla-it/ctms-api/commit/2036559ded34d78d2c2418840e8362df46a06ce8#diff-1a390c7b8c5ef6d2a2186016146f3b93ce3a2c04874ab1bfe02e8f67f350e8bfR29)


- jwhitlock reported [the issue linked in this PR](https://github.com/mozilla-it/ctms-api/issues/238) on Jun 1, 2021

I don't know how / when `bsieber-mozilla`'s migrations were applied in prod, but maybe the first version of the migration was applied and ran, but the edit wasn't picked up 🤔. Pure speculation.

grahamalama (commented):
But in any case, some thoughts / questions:

- is it true that we have an index `idx_email_primary_email_lower`, and now we're adding this `idx_email_primary_unique_email_lower`? Maybe we can drop `idx_email_primary_email_lower` since `idx_email_primary_unique_email_lower` is that plus uniqueness

- if we do need this index, can we include a comment here explaining why, given the confusing history I pointed out in my previous comment?

- Can we reflect this index back onto [the model](https://github.com/mozilla-it/ctms-api/blob/2201ec68cc0c0a84ffdda84184bf130bd839cb37/ctms/models.py#L94)?

leplatrem:
You are perfectly right! I was only diffs between schemas, and didn't spot that we also have `idx_email_primary_email_lower`.

Stage/Prod have this additional one `idx_email_primary_unique_email_lower`.

I propose that we keep the one with uniqueness, and drop the other one. Let's see how I can rework the migration for that.

Thank you for being so thorough, as usual 🙏

ctms/models.py
```diff
@@ -93,7 +93,11 @@ def primary_email_insensitive_comparator(cls):
     # Indexes
     __table_args__ = (
         Index("bulk_read_index", "update_timestamp", "email_id"),
-        Index("idx_email_primary_email_lower", func.lower(primary_email)),
+        Index(
+            "idx_email_primary_unique_email_lower",
+            func.lower(primary_email),
+            unique=True,
+        ),
```
grahamalama, line 100: In this same way that this index "covers" `idx_email_primary_email_lower`, is there also a plain unique index on `primary_email` via the `unique` flag set in the [model](https://github.com/mozilla-it/ctms-api/blob/2201ec68cc0c0a84ffdda84184bf130bd839cb37/ctms/models.py#L60)? If not, maybe we can set that flag to `False`, since I think that's essentially there to generate a unique index.

grahamalama (approved):
Nice! One topic of investigation for indexes that I though of, but otherwise looks g2g and I'll throw the approval on now.

ctms/models.py
leplatrem: As discussed elsewhere:

> I would say that this index does not hurt. Only drawback of indexes is cost of write and db/ram size. I don't think we are the level of optimizing these

PR #994, opened 2024-10-28
Fix #742: Update integration tests to use recent basket image

### LINKED ISSUE #742: Upgrade integration tests to latest Basket
We pinned it in #738 and #741 

Changes upstream concern celery and migration to rq
https://github.com/mozmeao/basket/pull/1069

robhudson:
The big change since the last image was basket switched to using rq for a task queue.


leplatrem (approved):
👌

PR #997, opened 2024-11-12
Change codeowners from syseng to `robhudson`

grahamalama:


grahamalama:
@robhudson, who else should we add here, if anyone?

.github/CODEOWNERS
```diff
@@ -1,4 +1 @@
-# Thanks to @imbstack for the suggestion.
-# see http://code.v.igoro.us/posts/2019/07/codeowners.html
-
-* @mozilla-it/syseng
+* @robhudson
```
robhudson, line 1: I think we can also add `@mozmeao/backend` to include the 3 of us.

grahamalama: Will do!

grahamalama: Is that a team in https://github.com/orgs/mozilla-it/? All I was able to find was https://github.com/orgs/mozilla-it/teams/mozmeao. I don't think you can do cross-org codeownership like that, unfortunately

robhudson: Okay, makes sense. We could use the `@mozmeao` one. I think it would be nice to have a "main" but also a team as fallback.

robhudson:
Updated to use `* @mozilla-it/ctms-reviewers` to more easily manage.

robhudson (approved):

PR #1004, opened 2024-12-05
Create script for updating subscriptions in bulk

robhudson:
This was needed as we had some newsletter updates come through that didn't sync with CTMS. We exported those newsletter changes and ran with this script to update CTMS data.


pmac (approved):
📜

robhudson:
This was run. Closing as we don't need this but would like to keep this around for future reference.

PR #1019, opened 2025-01-28
Upgrade PostgreSQL to 15

leplatrem:


robhudson:
I looked at this locally. It fails intermittently.

The 500 `resp.content` is:
```
{"status":"error","checks":{"database":"error"},"details":{"database":{"status":"error","level":40,"messages":{"db.0002":"Contacts table empty"}}}}
```

The health check is counting the number of contacts and getting a `-1` which doesn't pass the [this assertion](https://github.com/mozilla-it/ctms-api/blob/main/ctms/routers/platform.py#L111), so the test fails.

There's a comment for the `count_total_contacts` function:
```
    Since the table is huge, we rely on the PostgreSQL internal
    catalog to retrieve an approximate size efficiently.
    This metadata is refreshed on `VACUUM` or `ANALYSIS` which
    is run regularly by default on our database instances.
```
Perhaps something changed slightly with those between Postgresql 12 and 15?

robhudson:
Updated with a fallback for cases where the table has not yet been vacuumed or analyzed.

PR #1022, opened 2025-02-27
Fix #1016: Add roles and permissions, plus CLI to manage them

### LINKED ISSUE #1016: Proposal: Adding Operation-Based Permissions to CTMS
### Background

Currently, the CTMS API uses client authentication credentials that provide blanket access to all operations. Any authenticated user can perform any action, leading to potential security and data integrity risks. To address this, I propose implementing operation-based permissions to assign specific access rights to specific named roles which are then assigned to the `api_client` records. Operations are loosely mapped to the CTMS API, e.g. `add_contact`.

### Objective

Introduce operation-based permissions to CTMS to
* Restrict access to operations based on roles.
* Ensure control over API operations (e.g., create, read, update, delete) at the operation level.
* Enhance the security of sensitive user data, particularly the "emails" table, by limiting delete access to a small set of roles.

#### Core Components

1. **Access Groups**
    - Define specific named roles.
    - Associate API credentials with one or more roles.
    - Include an overall `Admin` role similar to Django's superuser.
2. **Operation-Based Permissions**
    - Implement permissions for specific operations.
    - Example permissions:
        - `create_contact`
        - `read_contacts`
        - `update_concact`
        - `delete_contact` (restricted to a small subset of roles).
3. **Role-Based Access Control**
    - Map roles to permissions.
4. **User-Based Access Control**
    - Consider whether we also need user-based access control

#### Implementation Steps

1. **Database Schema Updates**
    - Add tables/models to track roles, permissions, and operation-level associations:
        - `roles`: Store role definitions (e.g., `AMO Admins`, `Newsletter Managers`).
        - `permissions`: Define actions for each type (e.g., `create`, `read`, etc.).
        - `role_permissions`: Map roles to permissions.
        - `client_permissions`: If a client doesn't fit into a specified role, they could have their own specific permissions assigned.
2. **Permission Checking Tooling**
    - Write tooling to help manage and view permissions and roles.
    - Develop permissions checks mechanism (via fastapi dependency injection?)

3. **API Updates**
    - Update existing endpoints to:
        - Enforce permission checks.
        - Respond with appropriate error messages (e.g., 403 Forbidden) for unauthorized actions.

robhudson:
See the `docs/access_controls.md` and `docs/cli.md` files here for a bit more details.

This doesn't currently add any permissions or permission dependencies to any routes, that will come later.


robhudson:
@leplatrem or @grahamalama: Curious if either of you would have time for a quick review? I know `+2,312` lines isn't quick but perhaps a high level review?

docs/access_controls.md
```diff
@@ -0,0 +1,114 @@
...
+
+This can be used to access the API, such as:
+
+```sh
+curl --oauth2-bearer <token> <server_prefix>/ctms?primary_email=<email>
+```
+
+## Protecting Endpoints in FastAPI
+
+To protect an API endpoint and **require a specific permission**, use the **`with_permission`**
```
leplatrem, line 88: Nit: I understand that in this phase, you want to document how the permissions will be used in the code base, but will readers of this doc be developers? Maybe we could just rephrase from *To protect* to *This is how API endpoints are protected*

docs/access_controls.md
```diff
@@ -0,0 +1,114 @@
...
+### Key Characteristics
+- Permissions are not assigned directly to API clients.
+- Roles are the only way permissions are granted to clients.
+- A single permission can be used in multiple roles.
+
+### Example Permissions
+| Permission Name | Description |
+|----------------|-------------|
+| `manage_contacts` | Grants the ability to create, edit, and delete contacts |
+| `view_updates` | Allows access to updated contacts |
```
leplatrem, line 26: nit: I don't find `view_updates` very easy to grasp. Maybe `delete_contact` is better?

ctms/permissions.py
```diff
@@ -0,0 +1,98 @@
...
+        .where(Permissions.name.in_(permission_names))
+        .where(Permissions.id == RolePermissions.permission_id)
+        .where(RolePermissions.role_id == ApiClientRoles.role_id)
+        .where(ApiClientRoles.api_client_id == api_client_id)
+    ).scalar()
+
+    if has_perm:
+        return True
+
+    return False
```
leplatrem, line 72: nit: above you used `bool(has_perm)`

leplatrem (approved):
I'm sorry Rob for the delay.

This is great! Super clean 👌 

I have one question about the permissions. Since permissions are meant to be used in code, do you think it's relevant to expose their manipulation to admins? Intuitively I would either create them on the fly when a view refers to it, and put their creation in migration code so that they get created automatically when a new version is deployed. What do you think?

docs/access_controls.md
robhudson: My intention is this is for the developers to more easily manage the permissions. But this is a good suggestion regardless.

robhudson:
> I have one question about the permissions. Since permissions are meant to be used in code, do you think it's relevant to expose their manipulation to admins?

I think having a CLI to manage permissions and roles is valuable, as it provides flexibility for admins when needed. That said, I agree that the most common approach would be to add permissions via migrations alongside the code that uses them. This ensures they are always created when deploying a new version, keeping them in sync with the application logic.

docs/access_controls.md
robhudson: Yep, hard to come up with arbitrary examples. :) I was trying to show various levels of CRUD operations that might come up.

ctms/permissions.py
robhudson: I remember having to change this to appease the typing gods, but let me revisit it and see if that changed.

robhudson: Runs fine using `bool(has_perm)`. Thanks for catching the inconsistency. \o/

PR #1028, opened 2025-03-13
Update fallback health check query

robhudson:
This avoids a very expensive COUNT query which could bog down the database

### DIFF
--- a/ctms/crud.py
+++ b/ctms/crud.py
@@ -55,17 +55,21 @@ def ping(db: Session):
 def count_total_contacts(db: Session) -> int:
     """Return the total number of email records.
 
-    Since the table is huge, we rely on the PostgreSQL internal
-    catalog to retrieve an approximate size efficiently.
-    This metadata is refreshed on `VACUUM` or `ANALYSIS` which
-    is run regularly by default on our database instances.
+    Since the table is huge, we rely on the PostgreSQL internal catalog to retrieve an approximate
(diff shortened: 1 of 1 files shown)


bkochendorfer (approved):

ctms/crud.py
```diff
@@ -55,17 +55,21 @@ def ping(db: Session):
...
+    and return 1 to indicate its presence.
     """
-    result = db.execute(text("SELECT reltuples AS estimate " "FROM pg_class " f"where relname = '{Email.__tablename__}'")).scalar()
+    result = db.execute(text(f"SELECT reltuples AS estimate FROM pg_class WHERE relname = '{Email.__tablename__}'")).scalar()
+
     if result is None or result < 0:
-        # Fall back to a full count if the estimate is not available.
-        result = db.execute(text(f"SELECT COUNT(*) FROM {Email.__tablename__}")).scalar()
-    if result is None:
-        return -1
```
pmac, line 68: we no longer want to return `-1` in any case?

pmac (approved):
👾

ctms/crud.py
robhudson: For now this unblocks the postgresql upgrade. I will circle back around and discuss with SRE if we want this to 500 and cause pods to be unhealthy as a signal that the `ANALYZE` isn't being run.

